PR #59, opened 2015-11-17
Je rectangular install

joneit:


snackycracky:
me excited, wow such polymer removal ;)

snackycracky:
aaand its merged

PR #102, opened 2016-03-30
fixed aggregations ordering issues

stevewirts:
@joneit @Dwaynekj


Dwaynekj:
@stevewirts It might be a little cleaner to use [process.ENV](http://thewebivore.com/super-simple-environment-variables-node-js/) to set a path to our local dev hyper-analytics module or any module

PR #108, opened 2016-04-22
feature/GRID-183: Draft of getSelectionRows with hidden rows for in-memory data

Dwaynekj:
@joneit 
Test in 
```
Object.keys(g.getRowSelection()).length
Object.keys(g.getRowSelection(true)).length
```

### DIFF
--- a/src/Hypergrid.js
+++ b/src/Hypergrid.js
@@ -1651,10 +1651,13 @@ Hypergrid.prototype = {
     },
 
 
-    getRowSelection: function() {
+    getRowSelection: function(includeHidden) {
         var c, column, self = this,
             selectedRowIndexes = this.selectionModel.getSelectedRows(),
-            numCols = this.getColumnCount(),
+            numVisCols = this.getColumnCount(), //Would like to rename to getVisibleColumnCount but would break the examples
+            hiddenCols = this.getHiddenColumns(),
+            numHiddenCols = hiddenCols && hiddenCols.length,
+            numCols = (includeHidden) ? (numVisCols + numHiddenCols) : numVisCols,
             result = {};
 
         function setValue(selectedRowIndex, r) {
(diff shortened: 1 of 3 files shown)


src/behaviors/JSON.js
```diff
@@ -83,6 +83,14 @@ var JSON = Local.extend('behaviors.JSON', {
     getHeaders: function() {
         return this.dataModel.getHeaders();
     },
+    /**
+     * @memberOf behaviors.JSON.prototype
+     * @return {string} The field at `colIndex`.
+     * @param {number} colIndex - the column index of interest
+     */
+    getField: function(colIndex) {
+        return colIndex === -1 ? 'tree' : this.getColumnFromFullList(colIndex).getField();
```
Dwaynekj: This has performance implications as anyone calling this will not only be searching the visible list of columns

src/dialogs/ColumnPicker.js
```diff
@@ -82,7 +82,8 @@ var ColumnPicker = Dialog.extend('ColumnPicker', {
         if (this.visibleColumns) {
             var columns = this.behavior.columns,
                 tree = columns[0];
-
+            //breaking encapsulation
+            //Should be using setters and getters on the behavior
```
Dwaynekj: Why is the columnpicker directly editing the behavior's state?

src/Hypergrid.js
```diff
@@ -1670,6 +1673,11 @@ Hypergrid.prototype = {
         return result;
     },
 
+    getHiddenColumns: function(){
+        //A non in-memory behavior will be more troublesome
+      return this.behavior.getHiddenColumns();
```
Dwaynekj: I'm not sure if this should be exposed on the Grid object as remote behavior wouldn't want access to call ALL hidden columns

Dwaynekj:
@joneit Is this published? Should we tell william to test?

PR #113, opened 2016-05-09
Feature/Grid-184

Dwaynekj:


joneit:
DJ, how are your "synthetic" events called? (Where are they called from?) Where's the 'onchange' event?

joneit:
DJ, I don't think we need to add `Hypergrid.prototype.hMax` ... when would it differ from `this.sbHScroller.range.max` (which has several references)?

Also, please change the links in the jsdoc comments from:

[fin-vampire-bar](http://datamadic.github.io/fin-vampire-bar/components/fin-vampire-bar/)

to

{@link http://openfin.github.io/finbars/FinBar.html|FinBar}

or 

{@link https://github.com/openfin/finbars|FinBar}

joneit:
Regarding the "lingering repaint" commit, let's move the `behavior.changed()` call to below the `else` block. thx

Dwaynekj:
@joneit Ready for review

Dwaynekj:
@joneit Is there a way for list dragon to notify when a drop happens so that we can be more intelligent when firing 'column-changed'

Dwaynekj:
We can use https://github.com/openfin/fin-hypergrid/blob/feature/big-refactor/src/behaviors/Behavior.js#L1181
for updating the column indices. It's not use in currently so I can modify it

Dwaynekj:
The repaint call has been resolved in another PR

Dwaynekj:
I've update the synthetic events to only declare the ones actually being called internally

Dwaynekj:
Switched to `this.sbHScroller.range.max`

Dwaynekj:
@joneit 
#API Test#
```
g.behavior.getPrivateState().columnIndexes
g.behavior.setColumnIndices([1,2,3])
g.behavior.getPrivateState().columnIndexes
```

#Drag Spec#
- Dragging a column but returning it to its original position *will not* fire columnChangedEvent
- Dragging a column to a new position *will* fire columnChangedEvent

#Column Picker#

PR #140, opened 2016-05-31
IMPRV/GRID-233: Cell Rendering Docs & Examples

Dwaynekj:
@joneit


src/cellRenderers/CellProvider.js
```diff
@@ -234,10 +245,23 @@ var CellProvider = Base.extend('CellProvider', {
         config.minWidth = config.minWidth + 2 * (iconWidth);
     },
 
-    renderMultiLineText: function(gc, x, y, height, width, config, val) {
+    /**
+     * @summary Renders multiline text
+     * @param {CanvasGraphicsContext} gc
+     * @param {number} config.bounds.x - the x screen coordinate of my origin
```
joneit: Since `config.bounds` is a `Rectangle` just say:
```javascript
/**
 * @param {Rectangle} config.bounds - The bounding rect of the cell to render.
 */
```
Note however that Rectangle is not in this project. (It is in the "rectangular" npm module). So in this case, add it to ./src/jsdocs/external-types.js as a @typedef with a @see link to it's jsdoc page (http://openfin.github.io/rectangular/Rectangle.html). This will make Rectangle a known type to jsdoc throughout the app.

src/cellRenderers/CellProvider.js
```diff
@@ -234,10 +245,23 @@ var CellProvider = Base.extend('CellProvider', {
...
+     * @param {number} config.bounds.y - the y screen coordinate of my origin
```
joneit: I think `rect` makes more sense than `bounds`.

src/cellRenderers/CellProvider.js
```diff
@@ -234,10 +245,23 @@ var CellProvider = Base.extend('CellProvider', {
...
+     * @param {number} config.bounds.width - the width I'm allowed to draw within
```
joneit: Please use whole sentences with initial capital letter and final period. I know this sounds peevish but in fact it otherwise looks lousy on the rendered jsdocs page when you have what appear to be random sentence fragments all over the place.

src/behaviors/Behavior.js
```diff
@@ -6,7 +6,7 @@ var Base = require('../lib/Base');
 
 var Column = require('./Column');
 var dialogs = require('../dialogs');
-var CellProvider = require('../lib/CellProvider');
+var CellProvider = require('../cellRenderers/CellProvider');
```
joneit: If you intention in moving this to a new `cellRenderers` folder was to break out the various cell renderer functions into separate files, please go ahead and do so.

Dwaynekj: Yes but as a placeholder. No need to do it now.

joneit: As it will reorganize the way cell renderers are defined, I think for the better, it might be worth doing now. When are you thinking we would do it?

Dwaynekj: That the location of cellrenders would be a little clearer. CellProvider in lib doesn't strike as the first place we are calling cell rendering. This is also consistent with celleditor structure. Even more so many of files in lib should be broken out into utils and lib but I didn't want to go overboard

joneit: I completely agree and that is what I meant by "reorganize." Like the cellEditors folder, we would have a base class, etc.

The current CellProvider code should be made a base class which we should rename `CellRenderer`. Then each cell renderer should extend that class. Any currently "inner" functions needed by more than one renderer should be moved to the prototype.

This would let the app developer "require" only the renderers he needs. Which is cool.

We also need to get away form the index.js thing where we "pre-register" all of them just so they can be randomly referenced by name later. Rather, we should force app developer to "require" only the ones he's using, and register them himself.

When do you think we should do this if not now? I am only saying this because looks like we're putting them off till end of week anyway so _we have the time._ AND I'd hate to see William start making cell renderers to the existing structure only to have to refactor his work in a few weeks. Although not a biggie, I suppose. It would be nice to smooth out the remaining wrinkles, like the crazy structures here. (Like, what does "cellCache" mean anyway?)

Dwaynekj:
@joneit ready for review

PR #151, opened 2016-06-05
moved `getCell` to `DataModel` and added `declaredRendererName` param

joneit:


joneit:
@Dwaynekj please look this over to see what I did and if anything stands out to you. seems to work fine.

PR #173, opened 2016-06-21
Imprv/cell editor cruft cleanout

joneit:
includes a couple of changes suggested by Abhang's question about how to programmatically start cell editing from keyup event


Dwaynekj:
👍

Issue #181, opened 2016-06-23
remove selectAll() on edit focus?


RaghavPro:
When you try to edit something, the focus event gets fired which calls a custom `selectAll` function.

with 'type to edit' as you go on typing it ends up overwriting the first character because it was selected. 
so for example, if you enter `fin-hypergrid` in a cell it becomes `in-hypergrid`

Dwaynekj:
@RaghavPro Thanks for the PR. We will evaluate it as apart of current sprint

RaghavPro:
Hi. I'm planning to use this and would love to contribute.

Let me know if the PR need any changes.



On Fri, Jun 24, 2016 at 9:33 PM +0530, "Dwayne Jones" <notifications@github.com<mailto:notifications@github.com>> wrote:


@RaghavPro<https://github.com/RaghavPro> Thanks for the PR. We will evaluate it as apart of current sprint

-

Dwaynekj:
Will do. We will be reviewing the code this week

Dwaynekj:
Hey @RaghavPro Would you mind creating a quick 1min demo movie demonstrating issue (using standard video formats that quicktime or a VLC player could run)? Alternatively we can do a quick 5 min screenshare. We are EST

RaghavPro:
Screen share sounds good. I'm free right now if you are.

Dwaynekj:
Send a skype invite to dj_openfin

Dwaynekj:
Update: No party can reproduce. Closing issue

PR #182, opened 2016-06-23
Remove selectAll on focus.

RaghavPro:
See #181 

I think it would be best if you call `selectAll` only if it's a double click event.
I couldn't figure out how to do that.


Dwaynekj:
Will close unless the issue reemerges.

PR #193, opened 2016-07-12
GRID-272: Zoom issues, part 1

joneit:


joneit:
@Dwaynekj This just addresses the header text clip region error. Nothing to do with zoom _per se._ Still awaiting William's screen shots on zoom issues.

Dwaynekj:
👍

PR #208, opened 2016-07-20
make prototype feature methods @private

joneit:
This was merely to suppress references to these on every feature doc page (due to @extends)


Dwaynekj:
👍

PR #211, opened 2016-07-20
GRID-235: Externalize hyper-analytics (first step)

joneit:
dataModel/JSON.js:
- Now functions without aggregator or filter data sources.
- setData now re-uses previous data source if dataSource (1st param) is omitted.
- setData now passes 2nd param to constructor of first data source in the pipeline.

behaviors/Behavior.js, behaviors/JSON.js:
Changed calling signature for constructor (initialize method) from (grid, schema, dataRows) to (grid, dataRows, options).
Added options.fields, passed to dataMode/JSON.setData as 2nd param.

### DIFF
--- a/src/Hypergrid.js
+++ b/src/Hypergrid.js
@@ -63,9 +63,8 @@ function Hypergrid(div, options) {
     this.renderOverridesCache = {};
 
     options = options || {};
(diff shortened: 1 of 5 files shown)


Dwaynekj:
@joneit Will consume after I get aggregations to work again

PR #214, opened 2016-07-22
fix: fixes TypeError due to accessing out-of-bounds columns

ColinEberhardt:
Currently if you have a grid that has empty space to the right of the columns, the following error occurs on double-click:

```
Hypergrid.js:1261 Uncaught TypeError: Cannot read property 'getProperties' of undefinededitAt @ Hypergrid.js:1261
```

This fixes the issue by ensuring that the `editPoint` column is within the expected range

### DIFF
--- a/src/Hypergrid.js
+++ b/src/Hypergrid.js
@@ -1257,7 +1257,7 @@ Hypergrid.prototype = {
 
         this.stopEditing(); //other editor is open, close it first
 
-        if (editPoint.x >= 0 && editPoint.y >= 0) {
+        if (editPoint.x >= 0 && editPoint.y >= 0 && editPoint.x < this.getColumnCount()) {
             var editable = this.behavior.getActiveColumn(editPoint.x).getProperties().editable;
             if (editable || this.isFilterRow(editPoint.y)) {
                 this.setMouseDown(editPoint);


ColinEberhardt:
You can actually see this issue on your demo site by double-clicking to the right of the grid: https://openfin.github.io/fin-hypergrid/

Dwaynekj:
@ColinEberhardt Good catch. I'll test your branch, but again can you open a PR to develop

Dwaynekj:
@joneit @ColinEberhardt Tested! Works. We can include this in 1.0.8 when the PR to develop is opened

joneit:
@ColinEberhardt Thank you Colin for your contribution. Please do not trouble yourself with resubmitting to develop. I have already incorporated your logic into another branch. The reason I did this was: I noticed in testing your fix that there was a similar but separate problem when clicking to right of header cell. I put both fixes in [a new PR](https://github.com/openfin/fin-hypergrid/pull/219) (to develop) which slightly different logic in both code paths. I am closing this PR. Thanks again and we hope you will not be discouraged and contribute again in the future. Jonathan

ColinEberhardt:
Thanks @joneit @Dwaynekj - as this is a bug that effects your demo and likely quite a few users, then why not also merge to master and create a patch release? That way new users will not continue to encounter this bug?

Dwaynekj:
We are on the verge of publishing a new version of Hypergrid any way

PR #215, opened 2016-07-22
fix: removed README image from bundle

ColinEberhardt:
Saves 45KBytes of bloat ;-)


Dwaynekj:
@ColinEberhardt Thanks!! However can we merge this to develop. I'm aware its just the readme, but we don't want the community of getting into the habit of sending PR's to master

ColinEberhardt:
>  However can we merge this to develop. I'm aware its just the readme, but we don't want the community of getting into the habit of sending PR's to master

If you want people to PR against `develop` then perhaps you should set this as the default branch?

Dwaynekj:
The Readme/API on develop won't necessarily reflects the module released on [NPM](https://www.npmjs.com/package/fin-hypergrid)

PR #221, opened 2016-07-23
Don't bundle README images

ColinEberhardt:
replaces #215

PR #222, opened 2016-07-23
fix: removed README image from bundle

ColinEberhardt:
Saves 45KBytes of bloat ;-)

Replaces #221 #215


ColinEberhardt:
That's better ;-)

PR #232, opened 2016-08-01
Move readme image into subdir excluded from gulp build. Fixes #231.

### LINKED ISSUE #231: Useless image increases build size
The file `images/Hyperblotter Tabled Reduced Rows.png` gets built by Gulp for example into `demo/build/fin-hypergrid.min.js` and the distribution package at `https://openfin.github.io/fin-hypergrid/build/fin-hypergrid.min.js`. Removing the png file shrinks the minified build from 395407 to 335683 bytes (15% space savings).

jjrv:


jjrv:
Never mind, it seems this is already fixed in the `develop` branch.

PR #234, opened 2016-08-02
GRID-301: Treeview rows don't expand

joneit:


Dwaynekj:
👍 @joneit Missing some returns?

joneit:
They weren't "missing" per se. I added them to get info back from data calls. The feature chain "traditionally" just handled events it knew about else passed them on. I changed that for CellClick.js's `handleClick` "event" so that it _also_ passes it on if there was nothing to handle. To do this, I had to make the data call return some info, which in this case was whether or not the click was in the tree column. (You should make sure your `test` method checks for that too.)

PR #237, opened 2016-08-04
GRID-303: Show drill-down controls in its own column

joneit:


joneit:
@Dwaynekj This is mostly sample code for WIlliam, with a couple of hypergrid tweaks to make it feasible.

src/features/ColumnSorting.js
```diff
@@ -32,8 +37,12 @@ var ColumnSorting = Feature.extend('ColumnSorting', {
      * @param {Object} event - the event details
      */
     handleMouseMove: function(grid, event) {
-        var y = event.gridCell.y;
-        if (this.isFixedRow(grid, event) && y < 1) {
+        var gridCell = event.gridCell;
+        if (
```
Dwaynekj: What the purpose for this?

joneit: This code originally changed the mouse pointer to a "pointer" cursor (index finger) when hovering over a header cell (header row is currently always row 0, aka < 1). As I showed you, I added a line to this code to also check that the column's `unsortable` (new) property is falsy.

Dwaynekj:
👍

PR #261, opened 2016-08-09
IMPRV/minor cleanup

Dwaynekj:
@joneit 

Created an assets folder that holds the html, css, and images folder


joneit:
@Dwaynekj It's just three folders, hardly a mess. I really would prefer to leave as is.

Also, "assets" implies pieces called for from markup which isn't really what's going on here (with exception of readme).

PR #264, opened 2016-08-09
GRID-209: Uncheck rows after check all

joneit:


joneit:
This really only just solves one of several row selection issues!

PR #278, opened 2016-08-17
GRID-313: Allow a hidden default sort column

joneit:


joneit:
This implements William's request to be able to set the (hidden) ID as the default sort column, to essentially group the data but without a visible group sorting.

joneit:
Sorry this PR got cluttered up with by-id stuff. Unintended. :(

src/behaviors/Behavior.js
```diff
@@ -1665,91 +1661,6 @@ var Behavior = Base.extend('Behavior', {
...
         return this.dataModel.getFilteredData();
     },
-
-    /**
-     *
-     * @param newDataRow
-     */
-    addDataRow: function(newDataRow) {
-        requeue.call(this, this.gridBehaviorChanged);
-        return this.dataModel.addRow(newDataRow);
```
Dwaynekj: This code is from the other PR right? You just continued off of that branch?

joneit: @Dwaynekj Actually no... There were no interdependencies so instead I separated the PRs so they both branched off of develop. But yes you are right, this deletion more properly belongs in the other PR. I will move it.

PR #281, opened 2016-08-19
added first set of unit tests

rdepena:
@joneit @Dwaynekj check it out, nothing super involved but a basic harness. 

Will be back for some selection model tests.


Dwaynekj:
Shape tests are a good start?

PR #288, opened 2016-08-22
FEATURE/GRID-292: Make `halign` a "render property" with default of 'center'

Dwaynekj:
@joneit @rdepena 

API Call to test
grid.behavior.setColumnProperties(1,{halign: 'right'});
Note there is no exemption for "headers" they are treated like regular cells. If there areå semantics needed to have separate alignment that can be specified from the application layer by `getCell`

Best to test on the non-pink demo like
`demo/example.html`

as the main demo has some overrides in `getCell` that sets a different alignment


src/defaults.js
```diff
@@ -364,6 +364,12 @@ module.exports = {
      * @instance
      */
     fixedColAlign: 'center',
+    /**
+     * @default
+     * @type {string}
+     * @instance
+     */
+    defaultColAlign: 'center',
```
joneit: I would go with simply `halign` but I see you're trying to parallel `fixedColAlign` so I could live with `colAlign` or `dataColAlign`... but really I prefer `halign` which is already a familiar prop in getCell config.

src/behaviors/Column.js
```diff
@@ -9,7 +9,8 @@ var propertyNames = [
     'name',
     'header',
     'calculator',
-    'type'
+    'type',
+    'halign'
```
joneit: Not here. Rather on the column properties object, which automatically defers to the grid properties object before deferring to the theme properties object before deferring (finally) to the default properties object.

src/behaviors/JSON.js
```diff
@@ -226,11 +227,16 @@ var JSON = Local.extend('behaviors.JSON', {
...
-        if (x === 0 && this.hasHierarchyColumn()) {
-            return 'left';
+        var align = this.getPrivateState().defaultColAlign;
+        if (x === -1) {
+            align = 'right';
+        } else if (x === 0 && this.hasHierarchyColumn()) {
+            align = 'left';
         } else {
-            return 'center';
+            align  = this.getActiveColumn(x).halign || align;
```
joneit: should be:
```javascript
align = this.getColumnProperties(x).halign;
```
and remove line 230's definition (keep declaration)

Dwaynekj:
Will make changes today

Dwaynekj:
@joneit @rdepena Made changes

Dwaynekj:
@joneit Ready for review

src/defaults.js
```diff
@@ -369,7 +369,7 @@ module.exports = {
      * @type {string}
      * @instance
      */
-    defaultColAlign: 'center',
+    colHAlign: 'center',
```
joneit: Remove "col" because it can also pertain the the grid as a whole.

src/behaviors/Behavior.js
```diff
@@ -185,6 +185,9 @@ var Behavior = Base.extend('Behavior', {
         var tableState = this.getPrivateState();
         var properties = Object.create(tableState);
 
+
+        properties.halign = tableState.colHAlign;
```
joneit: Don't do this. Just make the entry in Defaults.js "halign" and you're done.

src/behaviors/JSON.js
```diff
@@ -60,7 +60,8 @@ var JSON = Local.extend('behaviors.JSON', {
             var column = this.addColumn({
                 index: index,
                 header: header,
-                calculator: calculator
+                calculator: calculator,
+                halign: ''
```
joneit: Remove this line. This is not needed here and is not being used.

src/defaults.js
```diff
@@ -364,6 +364,12 @@ module.exports = {
      * @instance
      */
     fixedColAlign: 'center',
+    /**
+     * @default
+     * @type {string}
+     * @instance
+     */
+    colHAlign: 'center',
```
joneit: Remove "col" because it can also pertain the the grid as a whole. Just make it 'halign'.

Dwaynekj: I think this should stay. We should have a col and rowAlign

PR #289, opened 2016-08-22
Exclude grouping DnD lists from Picker when unavailable

joneit:
Group view broke the Picker when no grouping methods are available (i.e., not mixed in). This fix _skips_ adding the grouping DnD lists when the methods are unavailable.

NOTE: This does not restore group picker effectiveness, which apparently has not worked since GRID-204 was merged.

### DIFF
--- a/src/dialogs/ColumnPicker.js
+++ b/src/dialogs/ColumnPicker.js
@@ -22,8 +22,11 @@ var ColumnPicker = Dialog.extend('ColumnPicker', {
         this.grid = grid;
 
         if (behavior.isColumnReorderable()) {
-            // grab the lists from the behavior
-            if (behavior.setGroups){
+            // parse & add the drag-and-drop stylesheet addendum
+            var stylesheetAddendum = stylesheet.inject('list-dragon-addendum');
+
+            // grab the group lists from the behavior
+            if (behavior.setGroups) {
                 this.selectedGroups = {
                     title: 'Groups',
                     models: behavior.getGroups()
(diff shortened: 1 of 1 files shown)


Dwaynekj:
What is group picker effectiveness?

src/dialogs/ColumnPicker.js
```diff
@@ -22,8 +22,11 @@ var ColumnPicker = Dialog.extend('ColumnPicker', {
         this.grid = grid;
 
         if (behavior.isColumnReorderable()) {
-            // grab the lists from the behavior
-            if (behavior.setGroups){
+            // parse & add the drag-and-drop stylesheet addendum
+            var stylesheetAddendum = stylesheet.inject('list-dragon-addendum');
+
+            // grab the group lists from the behavior
+            if (behavior.setGroups) {
```
Dwaynekj: @joneit I think I need to hookup this method to your drop event (or rather close dialog). Also I can move this to the addons folder

Dwaynekj: I would add
```
+            //TEMPORARY WORKAROUND. Column Picker will become a add-on with a noted dependency on group view
+            if (behavior.getGroups && behavior.getGroups().length > 0) {
                 this.selectedGroups = {
```
If groupView has been mixed in but the groups have been set to `[]`

Dwaynekj:
Also I tested [here](https://github.com/Dwaynekj/fin-hypergrid/tree/joneit-fix-column-picker) the column picker works with grouping for most combinations the following can be called programmatically or implicitly through the picker

```
b.dataModel.setGroups([1])
b.dataModel.setGroups([2])
b.dataModel.setGroups([3])
b.dataModel.setGroups([2,3])
b.dataModel.setGroups([4])
b.dataModel.setGroups([3])
```


However I did find an silent error when I tried `b.dataModel.setGroups([3,4])`. 
When you open the picker to see its the grouping Height and Birth Date
```

src/dialogs/ColumnPicker.js
joneit: working around what exactly? because if groups is `[]` shouldn't you be able to set them here? (I realize you are also setting `getCell` when you check "grouping" checkbox, so maybe we add that code here?)

joneit:
By "effectiveness" I meant:

1. Picking group with the group picker did not actually work. It just blanked the columns.
2. Group picker no longer works with aggregates (because `getGroups` and `getAvailableGroups` missing from add-ons/aggregations-view.js).

src/dialogs/ColumnPicker.js
joneit: +1 making picker an add-on...
Column Picker should be an add-on as well. Should they be separate dialogs?

Dwaynekj:
I will add the bugs to the ticket I created and resolve them for the next sprint.

PR #293, opened 2016-08-23
GRID-307: Remove 200px kludge

joneit:
@Dwaynekj Vertical scroll bar now positioned correctly on right edge of canvas, as in for example group-view.html.


Dwaynekj:
Cool will test tomorrow

Dwaynekj:
@joneit  looks good except for one minor thing.

If you open [here](http://localhost:63342/fin-hypergrid/demo/group-view.html) (which should be serving from IntelliJ by default), when the horizontal scroll is flushed right is the last column clipped?

PR #294, opened 2016-08-24
IMPRV/GRID-258 - Exploration of a non-destructive setData

Dwaynekj:
See Simulated Updates [POC](https://github.com/openfin/Hypergrid-Remote-POC)

@joneit @rdepena 

### Datapipeline
setData now does not rebuild the pipeline (as we planned) but applyAnalytics still has to be called to preserve filtering/sorting/etc state on the new data insertion.  Apply will also ensure that filtering correctly identifies the number of rows as when its in the pipeline, it becomes the authority on that fact (and it will only rebuild its index on an apply call). 
See [here](https://github.com/openfin/hyper-analytics/blob/master/js/DataSourceGlobalFilter.js#L59) and [here](https://github.com/openfin/hyper-analytics/blob/master/js/DataSourceGlobalFilter.js#L44)

### Scrolling
Same behavior as current. If the same number of rows are passed to setData the vScroll is unmoved otherwise it moves proportionally the *same position* to the number rows passed in

### The New DataSource???
Right now the analytics pipeline is sort of like a sql views funneled through one another
and the grid proper takes the actual data from the head of the pipe
secondarily even the un-touched data (the original array passed to setData) is stored in Hyper-analytics (*HA*) so even if your doing no “analytics” the grid defers to HA
I changed it so now the in-memory data is now in hypergrid core and can be fetched without needing *HA*. The underlying data is now always the first in the pipe.

### Future of removing HyperAnalytics
Ther filtering/sorting/etc modules from HA can still be used *BUT* but it also means that if a HA module like sorting provided a new data set (instead of a view) then it could really just be a setData  call triggered on an event with no need for applyAnalytics


add-ons/tree-view.js
```diff
@@ -105,8 +99,8 @@ TreeView.prototype = {
         dataModel.addPipe({ type: 'DataSourceTreeview', test: isTreeview });
 
         if (amInstance) {
-            this.grid.behavior.setData();
-            this.dataSource = dataModel.sources.dataSource;
+            this.grid.behavior.setPipeline();
+            this.dataSource = dataModel.dataSource;
```
joneit: The original of this line was a nonsense line of code; remove it.

src/dataSources/JSON.js
```diff
@@ -0,0 +1,188 @@
+'use strict';
+
+var Null = require('./Null');
```
joneit: Let's get rid of Null altogether. We're not using it and we don't need it. Originally Steve had "undefined" dataSources set to null, but we're not doing that anymore. And with our plan to build pipelines to suit your app, it'll never happen.

src/dataSources/JSON.js
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * @constructor
+ */
+var JSON = Null.extend('JSON',  {
```
joneit: I don't like calling it JSON. Calling the behavior and dataModel JSON was inaccurate at best and not such a good choice. How about DataSource or RawDataSource or something more descriptive?

joneit: Using .extend here also makes this object extensible. Are we sure we want that? I know it was extensible in hyper-analytics, but in practice it was never extended from. Let's just make it a regular function with a prototype.

src/dataSources/Null.js
```diff
@@ -0,0 +1,65 @@
+'use strict';
+
+var Base = require('../lib/Base');
+
+var Null = Base.extend('Null', {
```
joneit: Not needed as noted above.

src/dataModels/JSON.js
```diff
@@ -313,63 +274,84 @@ var JSON = DataModel.extend('dataModels.JSON', {
...
+        this.pipeline.forEach(function(sources, pipe) {
             var DataSource = analytics[pipe.type];
 
             pipe.name = pipe.name || getDataSourceName(pipe.type);
 
-            if (index === 0 && pipe.name !== 'source') {
-                throw 'Expected pipeline to begin with source.';
-            }
-
             if (pipe.parent) {
```
joneit: We really don't need to support `parent` any more. This was just an accommodation to trying to have one immutable pipeline that served both the needs of aggreate data and regular data. Now that we have a constructed pipeline, we can get rid of this!

src/dataSources/JSON.js
```diff
@@ -0,0 +1,188 @@
...
+
+    initialize: function(data, fields, calculators) {
+        this.setData(data, fields, calculators);
```
joneit: Maybe we should not be setting data implicitly on instantiation, but only subsequently on setData. We can leave as is for now and revisit this on the "launch sequence" ticket.

joneit:
General comments:

I have no issue with losing dataModels/Default.js and dataModels/InMemory.js.

Likewise with Local.js and Remote.js. Steve put those in on his last day. He seemed to feel it was important.

As noted in line notes, why do we even need Null? I know you have always considered it the definition of the interface but it was never meant as such; it was actually used as a null data souirce. And anyway the DataSource has an identical interface.

Honestly, I don't know why we're moving it. Is this just to eradicate any reference to hyper-analytics? Note that this data source serves (served) as the reference interface for hyper-analytics modules. Were you planning on deleting it from there? Because that will leave no interface definition there.

I do like the idea of making it a permanent first datasource in the pipeline, but it was already exactly that, wasn't it?

Regarding your comment above "Future of removing HyperAnalytics" all you are describing is a pipeline with only 1 datasource in it. This should be the default pipeline. Which means we're already there.

Finally, in general please try to avoid moving code around within files. It makes examining diffs more challenging.

Dwaynekj:
@joneit @rdepena 

Why Null?

Its is unfortunately more descriptive of the pipeline interface than the concrete version. It has, for example "click" and "revealRow", nulled functions that are used by the dataModel when there is a "drilldown". They are not implemented nor needed by the raw "plain" dataSource.

Why move it?

If we are to drop HyperAnalytics then we need the grid to take ownership of the data its given, no? HA should keep its version, though redundant, as it may actually be used outside of Hypergrid somewhere as pivot engine. Regardless, Hypergrid needs to move in the direction of a lesser coupling to it; especially for the simple fact that prior to these changes it couldn't even present the original data without HA. Now if we remove our opinions on sort and filtering (as we plan), the gird *literally* does not need HA.

Redundancy in permanently first in pipeline? 

Yes. At first glance the notion of 'rawDataSource' being first is exactly the same as before. However, If we are to further the **possibility** of setData + events as the **primary** engine for the user to supply a sort/etc then we can presumably remove the pipeline notion all together if needed. There would be no need for running a loop to set a pipeline. The only pipe would be the raw Data meant as a buffer for the user to edit and swap in. If we do decide to stick with a data transformation pipeline, then no harm done.

Regarding your comment above "Future of removing HyperAnalytics" 
We are not there. We can't remove HA if the grid defers reading/writing to the rawData to HA. If we remove HA right this moment on develop with no drop in replacement implementing the data retrieval the grid will not work.This is a step towards decoupling. See my first point.

Moving code around within files 
Sorry
I was getting annoyed with the order of how the methods were listed and that I needed to jump (split view) to their definitions quicker. The methods are not listed alphabetically nor grouped logically for quickly glance of a functionality/feature. Luckily jsdocs does a alphabetically sort by default, but thats in the docs

add-ons/tree-view.js
Dwaynekj: Will do

src/dataSources/JSON.js
Dwaynekj: See long answer below.

src/dataSources/JSON.js
Dwaynekj: Ok 2x

src/dataSources/Null.js
Dwaynekj: Ok

src/dataModels/JSON.js
Dwaynekj: Ok

src/dataSources/JSON.js
Dwaynekj: Yes currently the init of Behavior and DM calls setData. I need to fix those first

src/dataSources/JSON.js
```diff
@@ -0,0 +1,188 @@
...
+     */
+    getHeaders: function() {
+        return (
+            this.headers = /*this.headers || (temporary until I get bootstrapping running unopionated)*/ this.getDefaultHeaders().map(function(each) {
+                    return capitalize(each);
+                })
+        );
+    },
+
+    getCalculators: function() {
```
joneit: Remove getCalculators; no longer used.

src/dataSources/JSON.js
```diff
@@ -0,0 +1,188 @@
...
+    /**
+     * @memberOf DataSource.prototype
+     * @param y
+     * @returns {object[]}
+     */
+    getRow: function(y) {
+        return this.data[y];
+    },
+
+    findRow: function(columnName, value) {
```
joneit: This version of `findRow` is out of date. Please paste in the version from hyper-analytics/js/DataSource.js.

src/dataSources/Null.js
```diff
@@ -0,0 +1,65 @@
...
+    isNullObject: function() {
+        return true;
+    },
+    getProperty: function getProperty(propName) {
```
joneit: Copy in the other methods from hyper-analytics/js/Base.js so this file can serve as a base for any data source.

src/dataSources/Null.js
```diff
@@ -0,0 +1,65 @@
...
+            return getProperty.call(this.dataSource, propName);
+        }
+    },
+    getDataIndex: function(y) {
+        return 0;
+    },
+    getRow: function() {
+        return null;
+    },
+    findRow: function(columnName, value) {
```
joneit: @Dwaynekj Unfortunately having this null method defined here (just to serve as an "interface definition") is problematic -- if this file is to serve as a base, as we discussed. The problem is the forwarding routines _depend on such methods being undefined_ and if they are defined here it will break the forwarding mechanism.

To be clear, by "forwarding routines" I mean this (from hyper-analytics/js/Base.js):

```javascript
    findRow: function() {
        if (this.dataSource) {
            return this.dataSource.findRow.apply(this.dataSource, arguments);
        }
    },
```

We could resolve this issue, although not performant, like this:

```javascript
    findRow: function() {
        if (this.dataSource && this.dataSource.findRow === this.findRow) {
            return this.dataSource.findRow.apply(this.dataSource, arguments);
        }
    },
```

Simpler and more performant would be to comment out all the null methods, or at least the ones for which an implementation in the data source "class" is optional. These would include all the ones for which there are forwarding functions. Obviously, methods _required_ to be implemented can be left uncommented _though of course they will never be called_ so might as well comment them out as well. We can solve this probelm simply by not depending on code for this purpose (interface definition). Just leave them as comments or better yet IMHO put them in an external document and link to that document.

joneit:
Remove ` || 'source' ` from [here](https://github.com/openfin/fin-hypergrid/pull/294/files#diff-f74f4d3ec8669c829ca4cdfc2ca391b6R853). No longer needed.

joneit:
Remove ` = this.dataSource` from [this line](https://github.com/openfin/fin-hypergrid/pull/294/files#diff-f3474484138f958e2d2bd1ab52d1a627R117):

was:
```javascript
    dataSource = this.dataSource = dataModel.sources.treeview,
```

now:
```javascript
    dataSource = dataModel.sources.treeview,
```

`this.dataSource` is not in use.

src/dataSources/JSON.js
Dwaynekj: Will rename as you requested

Dwaynekj: Will be kept as agreed

Dwaynekj:
Will the make the changes tomorrow

Dwaynekj:
Note to self. Move the applyAnalytics call in setData to the application layer. The user should know if they need an `apply` call

src/dataSources/JSON.js
Dwaynekj: @joneit Actually I can't recall our discussion here. Do you still want me to remove it?

src/dataSources/JSON.js
Dwaynekj: Done

src/dataSources/Null.js
Dwaynekj: Done

src/dataSources/JSON.js
Dwaynekj: Done

src/dataSources/JSON.js
Dwaynekj: Done

src/dataSources/Null.js
Dwaynekj: Done

Dwaynekj:
@joneit @rdepena Minor bugs reintroduced post code review changes. Closing PR temporarily

Dwaynekj:
@joneit @rdepena PR review changes. With the exception that removing the parent notion from the pipeline causes errors in aggregations and grouping

Dwaynekj:
@joneit @rdepena Ready for merge. Any final thoughts?

PR #295, opened 2016-08-25
GRID-274: Avoid formatting on type-to-edit

joneit:


Dwaynekj:
@joneit I don't have the confidence to review this. You can use your judgement here

PR #300, opened 2016-08-28
BUG/GRID-153

Dwaynekj:
@joneit 
Brings Idle CPU time down from 5% to <1%


joneit:
Cool... so was there a fincavas change that goes with this ticket? Or was this the whole problem?

Dwaynekj:
No changes in Fincanvas

PR #303, opened 2016-08-29
Version 1.0.10

joneit:


joneit:
Note that I already pushed the version-history.html changes from github.

joneit:
Updated README.md to point to [latest releases](https://github.com/openfin/fin-hypergrid/releases/latest) rather than [version-history.md](https://github.com/openfin/fin-hypergrid/blob/master/version-history.md).

Keeping version-history.md for now but updated it with a deprecation notice and a link to the latest release page.

PR #305, opened 2016-08-29
Revert "BUG/GRID-153"

joneit:
Reverting... needs more research. turning off paint loop just means that repaints are done immediately and not periodically. not sure why turning it off reduced idle-time CPU usage, possibly simply because requestAnimationFrame was no longer running. in any case it only helped at idle time.


Dwaynekj:
ok. Good to know

PR #313, opened 2016-08-31
IMPRV/GRID-265

Dwaynekj:
@joneit @rdepena 

package.json to reflect a minor breaking change as the behavior must now be explicitly set by user
setFilterRow now defaults to false
noDataMessage now defaults to ''

### The Grid can be initialized as 

```
var g = new Hypergrid()
g.setBehavior(fin.Hypergrid.behaviors.JSON); //Initial Paint
g.setContainer('#id') //optional. Default div#hypergrid either searched or created. Calling post setBehavior will simply move the current grid instance
g.setFormatter(/*...*/); //optional
g.setPipeline([p1,p2]); //optional
g.setData();
```
### OR 
```
var g = new Hypergrid('#id');
g.setBehavior(new fin.Hypergrid.behaviors.JSON(g));  //Initial Paint
```
### OR
```
var g = new Hypergrid('#id', {Behavior: fin.Hypergrid.behaviors.JSON});  //Initial Paint
```
### OR
```
var g = new Hypergrid({Behavior: fin.Hypergrid.behaviors.JSON, container: '#id'}); //Initial Paint
```

- setContainer 
> Dynamic and will re-append the existing canvas and scrollbars if they already exist
- setFormatter 
> Dynamic and not needed for grid initialization
- setData 
> Remains Dynamic, implicitly call with setBehavior will default to [] if no data passed
- setPipeline 
> Dependent on https://github.com/openfin/fin-hypergrid/pull/294 and will be Dynamic
- setBehavior 
> implicitly instantiate dataModel, pipeline, fincanvas and scrollbars. WILL RENDER THE GRID.

### DIFF
--- a/demo/basic.html
+++ b/demo/basic.html
@@ -8,11 +8,10 @@
         window.onload = function() {
             fin.Hypergrid.JSON.get('data/basic.json', function(data) {
                 var div = document.querySelector('div#json-example'),
-                    jsonGrid = new fin.Hypergrid(div, { data: data });
+                    jsonGrid = new fin.Hypergrid(div);
 
-                jsonGrid.addProperties({ showFilterRow: false });
-
-                jsonGrid.behavior.setData(data);
+                jsonGrid.setBehavior(new fin.Hypergrid.behaviors.JSON(jsonGrid));
+                jsonGrid.setData(data);
             });
         };
     </script>
--- a/demo/example.html
+++ b/demo/example.html
@@ -5,24 +5,22 @@
 </head>
 <body>
 
-    <div id="fin-grid"></div>
 
     <script src="build/fin-hypergrid.js"></script>
 
     <script>
-var data = [
-    { symbol: 'APPL', name: 'Apple Inc.', prevclose: 93.13 },
-    { symbol: 'MSFT', name: 'Microsoft Corporation', prevclose: 51.91 },
-    { symbol: 'TSLA', name: 'Tesla Motors Inc.', prevclose: 196.40 },
-    { symbol: 'IBM', name: 'International Business Machines Corp', prevclose: 155.35 }
-];
-
-var grid = new fin.Hypergrid('#fin-grid', { data: data });
-
-grid.addProperties({
-    showRowNumbers: false,
-    showFilterRow: false
-});
+        var data = [
+            { symbol: 'APPL', name: 'Apple Inc.', prevclose: 93.13 },
+            { symbol: 'MSFT', name: 'Microsoft Corporation', prevclose: 51.91 },
+            { symbol: 'TSLA', name: 'Tesla Motors Inc.', prevclose: 196.40 },
+            { symbol: 'IBM', name: 'International Business Machines Corp', prevclose: 155.35 }
+        ];
+        var grid = new fin.Hypergrid();
+        grid.setBehavior(new fin.Hypergrid.behaviors.JSON(grid));
+        grid.setData(data);
+        grid.addProperties({
+            showRowNumbers: false
+        });
     </script>
 </body>
 </html>
--- a/demo/js/aggregations-view.js
+++ b/demo/js/aggregations-view.js
@@ -25,6 +25,7 @@ window.onload = function() {
             lastPet: rollups.last(2),
             stdDevPets: rollups.stddev(2)
         },
+        options = { data:  window.people1 },
         groups = [5, 0, 1],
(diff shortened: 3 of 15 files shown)


joneit:
@Dwaynekj I love what you've done here! I did not attemp to run this since setPipeline is missing from this PR... but I went over the code carefully. Just a few questions/comments (in no particular order):

0. Re the new `Hypergrid.prototype.setBehavior`:
  * Let's overload the required `behavior` param so you can as an alternative just pass give constructor. I believe this will be 99.9% of the use cases. (So when it is of type function, instantiate it with `this.grid` as only parameter.)
  * Regarding the formal parameters `dataRows, options` here as well as in `setData`: Let's make `dataRows` omittable in case there's an `options.data` definition
    * Logic: use 2nd param if an array; otherwise assume 2nd param is options object and use `options.data`
0. We need to add `pipeline` as an option in keeping with our intention to allow the `options` object to be a full and useable alternative to the discrete calls.
0. Please confirm regarding supplying `data` to the constructor that once we resolve the fincanvas chicken-and-egg issue we will be able to clean up the examples of initializing with discrete calls (_i.e.,_ when no `options` object) — so we can remove `{ data: data }` from those examples.
0. _Please confirm_ regarding the sample code in your PR description above...
  * ...you meant `Hypergrid` rather than `Grid`
  * ...`data` remains a member of `options` (as coded in Hypergrid.js) and that you're not planning on changing it a separate additional parameter
  * ...in the last example (after //OR) you meant "without" rather than "with"
0. Same idea with Hypergrid constructor: Let's make 1st param `div` omittable in case there's an `options.container` (a new option)
  * Logic: use 1st param when of string type (a query selector) _or_ an instance of `HTMLElement`; otherwise assume 1st param is options object and use `options.container`
  * Also, let's rename the formal param `container` (rather than `div`)
0. When container _not_ given at all, let's look for a default container called 'hypergrid'. If that's not found, let's create a container (_à la_ D3, according to Kolba) and call it 'hypergrid' using `options.boundingRect` (a new option) if given or a default size and position if not given. This will further simplify the examples which I like very much.
0. Minor stuff I noticed:
  * **Hypergrid.js:**
    * Line 741 is a duplicate of line 740
    * Let's move `|| {}` from the `setFormatter` call to the body of the new function; and rename `loc` formal param to `options` (now that it won't conflict with the constructor's `options`).
    * Let's move `if (!grid.sbHScroller || !grid.sbVScroller)` to inside of `initScrollbars`.
    * `addFinEventListener` should have been removed long ago so let's just remove it entirely at this point. (The way you altered it won't work at all anyway; not sure what you had in mind there!)
  * **basic.html:** 
    * Missing comma end of line 11
0. What broke exactly? that we bumping minor version number?
0. I see you are introducing methods with leading underscore character. While I personally have no problem with that, we have not been doing that up to now so I just don't think we should suddenly start doing so now. Let's remove the underscores and just mark them `@private`. (Yes, there are a handful of existing methods with such naming, but SWirts used it to mean "daughter function" rather than "private" so really introducing this usage is confusing.)
0. Pre-existing issue: Wondering why `themeInitialized`, `gridTheme` (formerly `polymerTheme`), and `globalProperties`are shared vars rather than instance vars. As shared vars, they apply to all grids on the page and I'm thinking that is opinionated. (These seem to be the only shared vars.)

Dwaynekj:
JE #1 Done

Dwaynekj:
JE 2 Done

Dwaynekj:
JE-3 & JE-4 Discussed and fixed

Dwaynekj:
JE-8 & 9 Discussed

Dwaynekj:
JE-7 - 740 is shapeChanged while line 741 is stateChanged maybe one is not is use?

joneit:
Following up...
* JE-5/6: This is a really cool nice-to-have but if you haven't started this yet, let's just do it later as it is fully backwards compatible.
* JE-7
  * you are right, it is not a duplicate. leave as is.
  * please remove `addFinEventListener`
* JE-9
  * Please at remove leading underscores, as discussed.
  * We also discussed propagating the `Hypergrid` context with `.call(this` and change references in these inner functions from `grid` to `this` and remove the formal param? Not pretty, but we do this elsewhere (a lot); reduce the subtlety by documenting with jsdoc comment `@this {Hypergrid}`.

Dwaynekj:
JE-5,6,7,9 done

Dwaynekj:
All demos ok except for calculated columns.

PR #314, opened 2016-08-31
Fix column cell properties methods

joneit:
The row coordinate (`y`) was being deindexed but was assumed to be a data row coordinate when in fact it was a grid coordinate. I added code to check if it was in the grid range and subtract the number of header rows before deindexing and added it back afterwards.

This may become a non-issue once we normalize grid coordinates to 0 in GRID-283 which is in this sprint, so let's hold off merging this until we're sure we need it or if 283 is not finished in time.

### DIFF
--- a/demo/js/demo.js
+++ b/demo/js/demo.js
@@ -998,7 +998,9 @@ window.onload = function() {
 
         grid.setState(state);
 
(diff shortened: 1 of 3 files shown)


joneit:
Closing. Will make it GRID-283's first commit.

PR #317, opened 2016-09-03
cell props obj now created from column props obj

joneit:
Polishes off the way we build and handle the cell properties object, bringing it into line with the way the other properties objects work:

* The cell properties object is now `Object.create`'d from the column properties objects. This is in keeping with the way the column properties object is created from the grid properties object. Up to now, this cascade had to be done explicitly inline wherever it was needed.
* With this new paradigm, the renderer no longer needs to mix in the cell properties object as a final step after calling `getCell` but before calling the cell renderer's `paint` method. (However, added a new render property `reapplyCellProperties` to get the old behavior, where cell properties override `getCell` settings.)
* Cell properties objects no longer restricted to data cells and can now be used for header cells.
* Isolated all cell properties code into a separate (new) file, behavior/Cell.js. (This is actually just mixed into Columns.prototype by Columns.js, but separating it out should make writing tests for it much cleaner and simpler.)
* This solution does maintain, for efficiency reasons, the paradigm of _avoiding_ creating a cell properties object for every cell (unlike the column properties objects, which are guaranteed to exist for every column). They are creating only when actually needed.

### DIFF
--- a/demo/js/demo.js
+++ b/demo/js/demo.js
@@ -986,11 +986,14 @@ window.onload = function() {
 
         grid.setState(state);
 
-        behavior.setCellProperties(idx.HEIGHT, 16, {
+        // decorate height cell in row "17"
+        var rowOrdinal = 17;
+        behavior.setCellProperties(idx.HEIGHT, behavior.getHeaderRowCount() + (rowOrdinal - 1), {
             font: '10pt Tahoma',
             color: 'lightblue',
             backgroundColor: 'red',
-            halign: 'left'
+            halign: 'left',
+            reapplyCellProperties: true
         });
 
         grid.addProperties({
@@ -1084,6 +1087,7 @@ window.onload = function() {
         });
 
         behavior.setColumnProperties(idx.TRAVEL, {
+            halign: 'right',
             format: 'francs'
         });
 
--- a/src/Hypergrid.js
+++ b/src/Hypergrid.js
@@ -3071,9 +3071,6 @@ Hypergrid.prototype = {
     isColumnSelection: function() {
         return this.resolveProperty('columnSelection') === true;
     },
-    getComputedRow: function(y) {
-        return this.behavior.getComputedRow(y);
-    },
     isColumnAutosizing: function() {
         return this.resolveProperty('columnAutosizing') === true;
     },
@@ -3246,10 +3243,18 @@ Hypergrid.prototype = {
     newRectangle: function(x, y, width, height) {
         return new Rectangle(x, y, width, height);
     },
-    getFormattedValue: function(x, y) {
+
+    /**
+     * @param {number} c - grid cell coordinate
+     * @param {number} r - grid cell coordinate
+     * @returns {*}
+     */
+    getFormattedValue: function(c, r) {
+        var x = this.behavior.getActiveColumn(c),
+            y = r - this.getHeaderRowCount();
         return this.formatValue(
-            this.getColumnProperties(x).format,
-            this.getValue(x, y + this.getHeaderRowCount())
+            this.behavior.getCellProperty(x, y, 'format'),
+            this.getValue(c, r)
         );
     },
 
--- a/src/behaviors/Behavior.js
+++ b/src/behaviors/Behavior.js
@@ -19,10 +19,6 @@ var noExportProperties = [
     'treeColumnPropertiesColumnSelection',
 ];
 
-var isNull = {
-    isNull: true
-};
-
 /**
  * @constructor
  * @abstract
@@ -721,13 +717,13 @@ var Behavior = Base.extend('Behavior', {
      * @memberOf Behavior.prototype
      * @desc update the data at point x, y with value
      * @return The data.
-     * @param {number} x - x coordinate
-     * @param {number} y - y coordinate
+     * @param {number} c - Grid column coordinate.
+     * @param {number} r - Grid row coordinate.
      * @param {Object} value - the value to use
      */
-    setValue: function(x, y, value) {
-        var column = this.getActiveColumn(x);
-        return column && column.setValue(y, value);
+    setValue: function(c, r, value) {
+        var column = this.getActiveColumn(c);
+        return column && column.setValue(r, value);
     },
 
     getDataValue: function(x, y) {
@@ -737,27 +733,38 @@ var Behavior = Base.extend('Behavior', {
     setDataValue: function(x, y, value) {
         this.dataModel.setValue(x, y, value);
     },
+
     /**
      * @memberOf Behavior.prototype
-     * @desc First checks to see if something was overridden.
-     * @return {*} The value at x,y for the top left section of the hypergrid.
-     * @param {number} x - x coordinate
-     * @param {number} y - y coordinate
+     * This does not include the column properties.
+     * @return {object} The cell properties for the cell at x,y in the hypergrid.
+     * @param {number} x - data x coordinate
+     * @param {number} y - data y coordinate
      */
     getCellProperties: function(x, y) {
-        var column = this.allColumns[x];
+        var column = this.getColumn(x);
         return column && column.getCellProperties(y);
+    },
+    /**
+     * @memberOf Behavior.prototype
+     * @return {*} The value at x,y in the hypergrid.
+     * @param {number} x - data x coordinate
+     * @param {number} y - data y coordinate
+     */
+    getCellProperty: function(x, y, key) {
+        var column = this.getColumn(x);
+        return column && column.getCellProperty(y, key);
     },
 
     /**
      * @memberOf Behavior.prototype
      * @desc update the data at point x, y with value
-     * @param {number} x - x coordinate
-     * @param {number} y - y coordinate
+     * @param {number} x - data x coordinate
+     * @param {number} y - data y coordinate
      * @param {Object} value - the value to use
      */
     setCellProperties: function(x, y, value) {
-        var column = this.allColumns[x];
+        var column = this.getColumn(x);
         if (column) {
             column.setCellProperties(y, value);
         }
@@ -1113,25 +1120,20 @@ var Behavior = Base.extend('Behavior', {
 
     /**
      * @memberOf Behavior.prototype
-     * @return {Object} The properties for a specific column. These are used if no cell properties are specified.
+     * @return {Object} The properties for a specific column.
      * @param {index} columnIndex - the column index of interest
      */
-    getColumnProperties: function(columnIndex) {
-        var column = this.columns[columnIndex];
-        if (!column) {
-            return isNull;
-        }
-        var properties = column.getProperties(); //TODO: returns `null` on Hypergrid.prototype.reset();
-        if (!properties) {
-            return isNull;
-        }
-        return properties;
+    getColumnProperties: function(x) {
+        var column = this.columns[x];
+        return column && column.getProperties();
     },
 
-    setColumnProperties: function(columnIndex, properties) {
-        var column = this.allColumns[columnIndex];
-        var columnProperties = column.getProperties();
-        _(columnProperties).extendOwn(properties);
+    setColumnProperties: function(x, properties) {
+        var column = this.getColumn(x);
+        if (!column) {
+            throw 'Expected data column.';
+        }
+        _(column.getProperties()).extendOwn(properties);
         this.changed();
     },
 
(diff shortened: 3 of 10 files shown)


joneit:
changed my mind; making this a commit on my current working branch instead (missing a file here anyway)

demo/js/demo.js
```diff
@@ -986,11 +986,14 @@ window.onload = function() {
 
         grid.setState(state);
 
-        behavior.setCellProperties(idx.HEIGHT, 16, {
+        // decorate height cell in row "17"
+        var rowOrdinal = 17;
+        behavior.setCellProperties(idx.HEIGHT, behavior.getHeaderRowCount() + (rowOrdinal - 1), {
```
joneit: This calculation needed for now to accommodate the way cell props are now grid row addressed rather than data row addressed so that they can be used for all rows including headers and not just data. This nonsense will however fall away once we do the GRID-283 (Refactor vertical grid coordinate).

demo/js/demo.js
```diff
@@ -986,11 +986,14 @@ window.onload = function() {
...
             font: '10pt Tahoma',
             color: 'lightblue',
             backgroundColor: 'red',
-            halign: 'left'
+            halign: 'left',
+            reapplyCellProperties: true
```
joneit: Forces renderer to reapply cell props after calling `getCell`.

src/Hypergrid.js
```diff
@@ -3246,10 +3243,18 @@ Hypergrid.prototype = {
...
+     * @param {number} r - grid cell coordinate
+     * @returns {*}
+     */
+    getFormattedValue: function(c, r) {
+        var x = this.behavior.getActiveColumn(c),
+            y = r - this.getHeaderRowCount();
         return this.formatValue(
-            this.getColumnProperties(x).format,
-            this.getValue(x, y + this.getHeaderRowCount())
+            this.behavior.getCellProperty(x, y, 'format'),
```
joneit: Access the cell props obj _when it exists_ (and has column props obj as prototype); _or_ the column props obj when cell props obj does not exist.

src/behaviors/Behavior.js
```diff
@@ -721,13 +717,13 @@ var Behavior = Base.extend('Behavior', {
...
-     * @param {number} x - x coordinate
-     * @param {number} y - y coordinate
+     * @param {number} c - Grid column coordinate.
+     * @param {number} r - Grid row coordinate.
      * @param {Object} value - the value to use
      */
-    setValue: function(x, y, value) {
-        var column = this.getActiveColumn(x);
-        return column && column.setValue(y, value);
+    setValue: function(c, r, value) {
```
joneit: Let's gradually change `x` (data column index) to `c` (grid column index) as appropriate and/or `y` (data row index) to `r` (grid row index) as appropriate.

This will really help make the code more comprehensible. This is just a start...

Importnat note: These coordinate changes do _not_ necessarily go in tandem as some important methods (such as `getCell`) currently use `x` and `r`.

src/behaviors/Behavior.js
```diff
@@ -1113,39 +1120,34 @@ var Behavior = Base.extend('Behavior', {
...
-            return isNull;
-        }
-        var properties = column.getProperties(); //TODO: returns `null` on Hypergrid.prototype.reset();
-        if (!properties) {
-            return isNull;
-        }
-        return properties;
+    getColumnProperties: function(x) {
+        var column = this.columns[x];
+        return column && column.getProperties();
```
joneit: I eliminated this `isNull` nonsense. Just returns `undefined` now. This is really only consumed by renderer. However, we will list it in our "breaking changes" doc for `1.1.0`.

src/behaviors/Behavior.js
```diff
@@ -19,10 +19,6 @@ var noExportProperties = [
     'treeColumnPropertiesColumnSelection',
 ];
 
-var isNull = {
-    isNull: true
-};
```
joneit: I eliminated this isNull nonsense. Just returns undefined now. This is really only consumed by renderer. However, we will list it in our "breaking changes" doc for 1.1.0.

src/behaviors/Column.js
```diff
@@ -319,4 +307,6 @@ Column.prototype = {
     }
 };
 
+_(Column.prototype).extendOwn(require('./Cell'));
+
```
joneit: This is where/how Cell.js gets mixed into `Column.prototype`.

src/defaults.js
```diff
@@ -769,6 +769,13 @@ module.exports = {
      * @instance
      */
     enableContinuousRepaint: false,
+
+    /** @summary Reapply cell properties after `getCell`.
+     * @type {boolean}
+     * @default
+     * @instance
+     */
+    reapplyCellProperties: false
```
joneit: I made this an opt-out. Breaking change, but better this way as it makes renderer faster and anyway we are planning to flip all the -ins into -outs on this `1.1.0` release and document them in "breaking changes" release notes.

src/lib/Renderer.js
```diff
@@ -950,15 +950,15 @@ var Renderer = Base.extend('Renderer', {
 
         var grid = this.grid,
             behavior = grid.behavior,
-            baseProperties = behavior.getColumnProperties(c);
+            column = behavior.getActiveColumn(c),
+            cellProperties = column && column.getCellProperties(r),
+            baseProperties = cellProperties || column && column.getProperties();
```
joneit: So here we get the combined (cascaded) cell props object _when it exists_ and only when it doesn't do we fall back to the column props obj.

src/lib/Renderer.js
```diff
@@ -950,15 +950,15 @@ var Renderer = Base.extend('Renderer', {
...
 
-        if (baseProperties.isNull) {
+        if (!baseProperties) {
             return;
         }
 
-        var columnProperties = baseProperties,
```
joneit: These are identical refs so I eliminated one of them.

src/lib/Renderer.js
```diff
@@ -975,92 +975,89 @@ var Renderer = Base.extend('Renderer', {
             isCellSelectedInColumn = grid.isCellSelectedInColumn(c),
             isCellSelectedInRow = grid.isCellSelectedInRow(r),
             areAllRowsSelected = grid.areAllRowsSelected(),
-            cellProperties;
```
joneit: Renamed to `config` which is more descriptive and less confusing.

src/lib/Renderer.js
```diff
@@ -975,92 +975,89 @@ var Renderer = Base.extend('Renderer', {
...
-
-        behavior.cellPropertiesPrePaintNotification(cellProperties);
+        var cell = behavior.getCellRenderer(config, c, r);
 
-        //declarative cell properties
-        if (isGridRow && isGridColumn) {
-            var overrides = behavior.getCellProperties(column.index, r);
-            _(cellProperties).extendOwn(overrides);
+        if (config.reapplyCellProperties) {
+            _(config).extendOwn(cellProperties);
```
joneit: Two changes here:

0. We no longer restrict cell prop objs to grid cells only.
0. We now do this only on request because the cell props are already in `config`: Only when explicitly needed to override modifications made to `config` by `getCell`.

demo/js/demo.js
Dwaynekj: ok.

src/behaviors/Behavior.js
Dwaynekj: A little confused on why getCell uses x and r, but sure

src/behaviors/Behavior.js
Dwaynekj: Ok

src/Hypergrid.js
```diff
@@ -3071,9 +3071,6 @@ Hypergrid.prototype = {
     isColumnSelection: function() {
         return this.resolveProperty('columnSelection') === true;
     },
-    getComputedRow: function(y) {
-        return this.behavior.getComputedRow(y);
-    },
```
Dwaynekj: Shouldn't this be deprecated?

src/behaviors/Behavior.js
```diff
@@ -1113,39 +1120,34 @@ var Behavior = Base.extend('Behavior', {
...
     },
 
-    setColumnProperties: function(columnIndex, properties) {
-        var column = this.allColumns[columnIndex];
-        var columnProperties = column.getProperties();
-        _(columnProperties).extendOwn(properties);
+    setColumnProperties: function(x, properties) {
+        var column = this.getColumn(x);
+        if (!column) {
+            throw 'Expected data column.';
```
Dwaynekj: Maybe remove the word data as you can set properties for headers inside of the columns as well

src/behaviors/Behavior.js
```diff
@@ -1500,22 +1502,12 @@ var Behavior = Base.extend('Behavior', {
         return false;
     },
 
-    getRowContextFunction: function(selectedRows) {
-        return function() {
-            return null;
```
Dwaynekj: What was this used for?

src/behaviors/Behavior.js
```diff
@@ -1500,22 +1502,12 @@ var Behavior = Base.extend('Behavior', {
...
-    },
-
     getSelectionMatrixFunction: function(selectedRows) {
         return function() {
             return null;
         };
     },
 
-    getComputedRow: function(y) {
-        return this.dataModel.getComputedRow(y);
```
Dwaynekj: What was this used for?

src/dataModels/JSON.js
```diff
@@ -607,42 +614,20 @@ var JSON = DataModel.extend('dataModels.JSON', {
...
-        return this.deprecated('buildRow', 'dataSource', '1.0.8');
-    },
-
-    /**
-     * @memberOf dataModels.JSON.prototype
-     * @param {number} y
-     * @returns {object}
-     */
-    getComputedRow: function(y) {
-        var rcf = this.getRowContextFunction([y]);
```
Dwaynekj: Did you test to confirm that aggregations/grouping was not broken by doing this?

src/dataModels/JSON.js
```diff
@@ -607,42 +614,20 @@ var JSON = DataModel.extend('dataModels.JSON', {
...
-        var row = {};
-        for (var i = 0; i < fields.length; i++) {
-            var field = fields[i];
-            row[field] = rcf(field)[0];
-        }
-        return row;
-    },
 
-    getValueByField: function(fieldName, y) {
-        return this.deprecated('getValueByField', 'dataSource', '1.0.8');
```
Dwaynekj: We're *un*-deprecating this?

src/dataModels/JSON.js
```diff
@@ -607,42 +614,20 @@ var JSON = DataModel.extend('dataModels.JSON', {
...
 
-            return topTotals[y - (headerRowCount - topTotals.length)];
+        if (r < headerRowCount && !hasToptotals) {
+            return topTotals[r - (headerRowCount - topTotals.length)];
         }
-        return this.dataSource.getRow(y - headerRowCount);
-    },
-
-    buildRow: function(y) {
-        return this.deprecated('buildRow', 'dataSource', '1.0.8');
```
Dwaynekj: This was only recently deprecated. I know they're annoying to see, but for the future should we decide how long we want to keep them?

src/lib/Renderer.js
Dwaynekj: I see. That means we will have to update our cellRenderer tutorials to reflect this

src/lib/Renderer.js
Dwaynekj: Good call

src/behaviors/Behavior.js
joneit: I don't like it either but it's so fundamental I think that at this point it really would screw people up to change it. `r` is appropriate (we do need cell renderers for header row cells) but not sure why it is `x` and not `c`.

src/lib/Renderer.js
joneit: yes... there isn't much to say other than using `reapplyCellProperties` as needed. Can be set on grid, column, cell; but also can be returned by `getCell`.

src/Hypergrid.js
joneit: ok

src/behaviors/Behavior.js
joneit: OK... but actually not sure what you meant here about headers, but turns out that what I meant was wrong too: I was thinking of the -1 column (the row handles) ... but that column does in fact have a properties object so never mind. But sure I'll remove "data."

src/behaviors/Behavior.js
joneit: Looks like it was an early stab at "computed rows"

src/dataModels/JSON.js
joneit: yes

src/dataModels/JSON.js
joneit: no, we're deleting it. you didn't see [my comment](https://github.com/openfin/fin-hypergrid/commit/391d6ea35e16a5279092de87a81abf257c38ba84)...

PR #320, opened 2016-09-06
GRID-324: Decorate group headers

joneit:


Dwaynekj:
@joneit Appears to be mostly demo code. Looks fine.

PR #334, opened 2016-09-15
IMPR/GRID-201

Dwaynekj:


src/dataModels/JSON.js
```diff
@@ -579,37 +548,9 @@ var JSON = DataModel.extend('dataModels.JSON', {
...
-    },
-
-    /**
-     * @memberOf dataModels.JSON.prototype
      * @param index
      * @param returnAsString
+     * @desc Provides the unicode character used to denote visually if a column is a sorted state
      * @returns {*}
      */
     getSortImageForColumn: function(columnIndex) {
```
joneit: So long as we're doing this, let's do it: Let's move this out too. Put an if around usage (in `getHeaderRowValue`).

src/dataModels/JSON.js
```diff
@@ -537,39 +536,9 @@ var JSON = DataModel.extend('dataModels.JSON', {
...
-                this.applyAnalytics({columnSort: true});
-            }
-        }
-    },
-
     /**
      * @memberOf dataModels.JSON.prototype
+     * @desc returns the columns that currently sorted and their intended direction of the sort
      */
     getSortedColumnIndexes: function() {
```
joneit: So long as we're doing this, let's do it: Let's move this (and behavior version) out too. It won't be called by `applyAnalytics` if there's no sorting datasource.

src/Hypergrid.js
```diff
@@ -1550,6 +1550,20 @@ Hypergrid.prototype = {
         });
         this.canvas.dispatchEvent(clickEvent);
     },
+    /**
+     * @memberOf Hypergrid.prototype
+     * @param {number} c - grid column index.
+     * @param {string[]} keys
+     */
+    fireSyntheticColumnSortEvent: function(c, keys) {
```
joneit: Why not in mixin?

add-ons/hyper-sorter/index.js
```diff
@@ -0,0 +1,29 @@
...
+    });
+
+    function mixInTo(name, instance, mixin) {
+        var object = objects[name];
+        var prototype = object && object.prototype || Object.getPrototypeOf(instance);
+        _(prototype).extend(mixin);
+    }
+}
+
+Hypersorter.prototype = {
```
joneit: remove this

add-ons/hyper-sorter/index.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+var _ = require('object-iterators');
+
+function Hypersorter(grid, objects) {
+    this.grid = grid;
```
joneit: make this a simple function. doesn't need to be an object with no methods or props:
```javascript
module.exports = function(grid, objects) {
```

add-ons/hyper-sorter/index.js
```diff
@@ -0,0 +1,29 @@
...
+    objects = objects || {};
+
+    mixInTo('Hypergrid', grid, require('./mix-ins/grid'));
+    mixInTo('Behavior', grid.behavior, require('./mix-ins/behavior'));
+    mixInTo('Column', grid.behavior, require('./mix-ins/column'));
```
joneit: this won't work. following will work so long as setData has been called and there is at least one column:
```javascript
mixInTo('Column', grid.behavior.allColumns[0], require('./mix-ins/column'));
```
alternatively, if we have access to the `Column` constructor, perhaps by adding to `src/behaviors/index.js`:
```javascript
objects.Column = Object.getPrototypeOf(grid).constructor.behaviors.Column;
mixInTo('Column', undefined, require('./mix-ins/column'));
```

src/Hypergrid.js
Dwaynekj: We will need the grid to trigger the events. The application developer will listen to these events and decide to act (or not) and if acting means doing a setData or applyAnaltyics

add-ons/hyper-sorter/index.js
Dwaynekj: Sorry I wasn't in the middle of changing this. I should held off on the PR a little longer

add-ons/hyper-sorter/index.js
Dwaynekj: ok

src/dataModels/JSON.js
Dwaynekj: In terms of Filtering UI and Sort UI, your saying leave that up to the user?

src/dataModels/JSON.js
Dwaynekj: this is the "state" that would be used for serialization

src/dataModels/JSON.js
joneit: All I was saying was if we're doing this mixin thing, let's not leave any random "sorting" methods in the core.

src/dataModels/JSON.js
joneit: yup just saying let's move it out

src/Hypergrid.js
joneit: this is fine, although the present task was just to moving sorting methods out, and in this case you have added one, so I'm just saying add it to the mix-in instead.

src/dataModels/JSON.js
joneit: yup is there some reason in cannot or should not be in the mixin file?

src/dataModels/JSON.js
joneit: no just saying move it out

src/dataModels/JSON.js
Dwaynekj: I was just thinking something like filter and sort state were agnostic or the dataSource/app logic -> setData called that handles it. So they wouldn't be apart of any mixin but the grid table state?

I wonder what filter state would look like? Maybe a string of the characters supplied for filtering for all columns

Dwaynekj: I thing we would need also a getFilteredColumnState. Do you understand what I mean as it being just agnostic grid UI state

src/dataModels/JSON.js
Dwaynekj: ok. That makes sense. If a user wants to customize the column headers to show column state they can certainly do that logic

src/Hypergrid.js
Dwaynekj: We need an API for the user to emit a sort and filtering events (fireSyntheticColumnFilterEvent or the like) as much as we need a cleaner event model. How they user chooses to respond to them was the point of the mixins, I thought.

Maybe you want to go with a model, where the user chooses their own event namespaces (outside of core UI events) and building their own model?

src/dataModels/JSON.js
Dwaynekj: Then again. Maybe we can just make the tableState object very generic and the user can use it as a central storage area for their custom application properties and we don't have any sort/filter state. We could make an area called `tablestate.customs = {}`. Thoughts?

add-ons/hyper-sorter/index.js
Dwaynekj: Done. Although I still `new` it in the pink demo

Dwaynekj:
Will merge, Since your PR is based on it https://github.com/openfin/fin-hypergrid/pull/336

joneit:
no, #336 is _not_ based on this...

PR #336, opened 2016-09-16
Additional global filter refactoring

joneit:


src/jsdoc/tutorials/filter-api.md
```diff
@@ -34,7 +34,7 @@ Instead of `FilterTree` in the above, you could use any object extended therefro
...
 
-The application programmer has full control of the type of filter to use and how it is set up. You can override the `getNewFilter()` method to instantiate a `DefaultFilter` with your own specific options; or you may instantiate a different filter object entirely. Alternatively, you can instantiate a new filter at any time and hand it to the grid by way of the {@link Hypergrid#setGlobalFilter} method.
+The application programmer has full control of the type of filter to use and how it is set up. You can override the `getNewFilter()` method to instantiate a `DefaultFilter` with your own specific options; or you may instantiate a different filter object entirely. Alternatively, you can instantiate a new filter at any time and hand it to the grid by way of the {@link Hypergrid#filter} setter.
```
Dwaynekj: this should be moved to github wiki as per @rdepena

src/cellEditors/FilterBox.js
```diff
@@ -82,7 +82,7 @@ var FilterBox = ComboBox.extend('FilterBox', {
             appendOptions: function(dropdown) {
                 if (!dropdown.length) {
                     // Various  operator options and/or optgroups vary per column based on `opMenu`.
-                    var opMenuGroups = this.grid.behavior.getGlobalFilter().opMenuGroups;
+                    var opMenuGroups = this.grid.behavior.filter.opMenuGroups;
```
Dwaynekj: opMenuGroups and parserCQL seem "domain" specific. You should note that this implementation of cell editor depends on extending filterTree if you wish to use this module. Otherwise the application develop should roll their own cell editor. Subsequently, we should also consider how this will affect properties like showFilterRow. It makes an assumption about where the row should be loaded (maybe too much special meaning).

The workflow I imagine is: I'm an application developer and I want to do filtering with a provided row for filtering.
It should be load Hyper-filter or my own filter plugin
Load this cell Editor or my own 
Override getCell to specify where to render the filterCell 
If they don't want a filterrow (and opting for a different UI) they won't get one


Lastly, we can remove logic from the grid that have to make special accomodations for a filter row
@rdepena Thoughts?

src/dataModels/JSON.js
```diff
@@ -174,10 +174,10 @@ var JSON = DataModel.extend('dataModels.JSON', {
                     value = at ? value.substr(0, at) + sortString + value.substr(at) : sortString + value;
                 }
             } else { // must be filter row
-                var behavior = this.grid.behavior,
-                    filter;
-                value = behavior.getGlobalFilter && (filter = behavior.getGlobalFilter()) &&
-                    filter.getColumnFilterState && filter.getColumnFilterState(this.getFields()[x]) || '';
+                if (!this.filter.getColumnFilterState) {
```
Dwaynekj: I know we are doing this to accomodate the current logic...
However, in the long run, this logic we've inherited seems too prescriptive. If they want to use filter they should specify it (as a opposed to us reading a defaults property), its cell renderer, and cell editor via some mechanism like `getCell` (since we don't have first class row objects due to perf reasons).

src/behaviors/Column.js
```diff
@@ -117,7 +117,7 @@ Column.prototype = {
         this._calculator = calculator;
 
         if (this.behavior.getGlobalFilter) {
-            this.behavior.getGlobalFilter().columnProperty(this.name, 'calculator', calculator);
+            this.behavior.filterProp(this.index, 'calculator', calculator);
```
Dwaynekj: Nice! This is definitely cleaner

package.json
```diff
@@ -18,15 +18,16 @@
   "dependencies": {
     "automat": "1.2.0",
     "chai": "^3.5.0",
-    "extend-me": "2.2.4",
-    "filter-tree": "0.3.39",
+    "extend-me": "^2.3",
+    "filter-tree": "0.3.40",
```
Dwaynekj: Shouldn't we able to remove this. Filter-tree is now the first of many implementations that can concur with the grid. However its not mandated.

I spoke to ricardo about mimicing @mrClapham's style of tutorial repos like so https://github.com/openfin/openfin-tutorial-hello-world

Dwaynekj:
PR rebased and merged to develop https://github.com/openfin/fin-hypergrid/pull/334

PR #337, opened 2016-09-16
GRID-380+381: Reconcile Column properties and move filter-tree method mop-up

joneit:


joneit:
Suggest you look at each commit separately as they represent entirely different issues.

Dwaynekj:
ok. With regards to the first commit. It looks very clean. Moving on to the next commit

PR #338, opened 2016-09-19
Added new test

Dwaynekj:
@joneit @rdepena


joneit:
re: shape, we haven't agreed on a "public" interface and we should only test for such methods. since javascript lacks public/private keywords, we could attempt to codify using jsdoc's @interface which I've been playing with and which works quite nicely....

PR #340, opened 2016-09-20
GRID-384: Plugins

joneit:
Got inspired on the train. Hope y'all like it.

### DIFF
--- a/demo/js/tree-view-separate-drill-down.js
+++ b/demo/js/tree-view-separate-drill-down.js
@@ -3,14 +3,10 @@
 
 'use strict';
 
-var grid, treeviewAPI;
+var grid;
 
 window.onload = function() {
     var Hypergrid = fin.Hypergrid,
-        Hyperfilter = Hypergrid.Hyperfilter,
-        Hypersorter = Hypergrid.Hypersorter,
-        drillDown = Hypergrid.drillDown,
-        TreeView = Hypergrid.TreeView,
         options = {
             // In order for the the State column not to sort the leaves (city names), we want it to use the "depth sorter" rather than the regular sorter used by the other columns. To make this work, the column does need to be called out when it differs from the tree (drill-down) column. In this demo, the tree column (containing the drill-down controls) is called name, which is the default. The `groupColumn` option normally defaults to whatever the `treeColumn` is set to but in this case we use it to call out the State column.
             groupColumn: 'State',
@@ -22,21 +18,15 @@ window.onload = function() {
     // Add a blank column.
     treeData.forEach(function(dataRow) { dataRow.name = ''; });
 
-    grid = new Hypergrid('div#tree-example');
-    grid.setData(treeData);
+    grid = new Hypergrid('div#tree-example', { data: treeData, plugins: [
+        Hypergrid.drillDown, // simple API install (plain object with `install` method) but no `name` defined so no ref is saved
+        Hypergrid.rowById, // ditto
+        Hypergrid.Hyperfilter, // object API instantiation; `$$CLASS_NAME` defined so ref saved in `grid.plugins.hyperfilter`
+        Hypergrid.Hypersorter, // object API instantiation to grid.plugins; no `name` or `$$CLASS_NAME` defined so no ref saved
+        [Hypergrid.TreeView, options] // object API instantiation with one arg; `$$CLASS_NAME` defined so ref saved in `grid.plugins.treeViewAPI`
+    ] });
 
-    // Install the sorter API (optional).
-    new Hypersorter(grid, { // eslint-disable-line no-new
-        Column: fin.Hypergrid.behaviors.Column
-    });
-
-    // Install the drill-down API (optional).
-    var dataModel = grid.behavior.dataModel,
-        dataModelPrototype = Object.getPrototypeOf(dataModel);
-    drillDown.mixInTo(dataModelPrototype);
-
-    var filterFactory = new Hyperfilter(grid);
-    grid.filter = filterFactory.create();
+    grid.filter = grid.plugins.hyperfilter.create();
 
     var idx = grid.behavior.columnEnum;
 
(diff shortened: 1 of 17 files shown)


demo/js/tree-view.js
```diff
@@ -22,21 +17,15 @@ window.onload = function() {
...
-    // Install the drill-down API (optional).
-    var dataModel = grid.behavior.dataModel,
-        dataModelPrototype = Object.getPrototypeOf(dataModel);
-    drillDown.mixInTo(dataModelPrototype);
+    grid.installPlugins([
+        Hypergrid.drillDown, // simple API install (plain object with `install` method) but no `name` defined so no ref is saved
+        Hypergrid.rowById, // ditto
+        Hypergrid.Hyperfilter, // object API instantiation; `$$CLASS_NAME` defined so ref saved in `grid.plugin.hyperfilter`
+        Hypergrid.Hypersorter, // object API instantiation to grid.plugin; no `name` or `$$CLASS_NAME` defined so no ref saved
+        ['treeviewAPI', Hypergrid.TreeView, options] // object API instantiation with one arg; [0] overrides any defined name so ref saved in `grid.plugin.treeViewAPI`
```
joneit: Maybe an object here? I usually use objects. The only reason I went with an array here is it "mimics" the constructor args. Alternative would look like:
```javascript
        { name: 'treeviewAPI', API: Hypergrid.TreeView, options, args: options }
```

joneit:
@Dwaynekj Please finish reviewing. Thanks.

Dwaynekj:
@joneit Working on sorting. I don't think we need to prioritize this for review (as it will take time and not originally apart of the sprint)

src/Hypergrid.js
```diff
@@ -263,6 +266,82 @@ Hypergrid.prototype = {
...
+     */
+    /**
+     * @desc Plugin installation:
+     * * Each simple API is installed by calling it's `install` method with `this` as first arg + any additional args listed in the `pluginSpec` (when it is an array).
+     * * Each object API is installed by instantiating it's constructor with `this` as first arg + any additional args listed in the `pluginSpec` (when it is an array).
+     *
+     * The resulting plain object or instantiated objects may be named by (in priority order):
+     * 1. if `pluginSpec` contains an array and first element is a string
+     * 2. object has a `name` property
+     * 3. object has a `$$CLASS_NAME` property
```
Dwaynekj: Should consider if extend-me specific properties will cause issues if we ever switch to ES6

src/Hypergrid.js
```diff
@@ -67,7 +67,7 @@ function Hypergrid(container, options) {
         options = container;
         container = null;
     }
-    this.options = options || {};
+    this.options = options = options || {};
```
Dwaynekj: huh?

src/Hypergrid.js
```diff
@@ -78,7 +78,7 @@ function Hypergrid(container, options) {
     this.numRows = 0;
     this.numColumns = 0;
     this.clearMouseDown();
-    this.setFormatter(this.options.localization);
+    this.setFormatter(options.localization);
```
Dwaynekj: I don't see the need of getting rid of the this keyword

demo/js/tree-view.js
Dwaynekj: I'd vote for object.
An object to me signifies all of the properties as having a relationship. An Array not so much

demo/js/tree-view-separate-drill-down.js
```diff
@@ -22,21 +18,15 @@ window.onload = function() {
     // Add a blank column.
     treeData.forEach(function(dataRow) { dataRow.name = ''; });
 
-    grid = new Hypergrid('div#tree-example');
-    grid.setData(treeData);
+    grid = new Hypergrid('div#tree-example', { data: treeData, plugins: [
+        Hypergrid.drillDown, // simple API install (plain object with `install` method) but no `name` defined so no ref is saved
+        Hypergrid.rowById, // ditto
+        Hypergrid.Hyperfilter, // object API instantiation; `$$CLASS_NAME` defined so ref saved in `grid.plugin.hyperfilter`
```
Dwaynekj: Definitely looks cleaner

src/Hypergrid.js
Dwaynekj: More importantly. We should also support an uninstall plugin

src/Hypergrid.js
```diff
@@ -263,6 +266,82 @@ Hypergrid.prototype = {
...
+     * One of:
+     * * simple API - a plain object with an `install` method
+     * * object API - an object constructor
+     * * array:
+     *    * first element is an optional name for the API or the newly instantiated object
+     *    * next element (or first element when not a string) is the simple or object API
+     *    * remaining arguments are optional arguments for the object constructor
+     * * falsy value such as `undefined` - ignored
+     *
+     * The API may have a `name` or `$$CLSSS_NAME` property.
```
Dwaynekj: Spelling Error $$CLSSS_NAME

src/Hypergrid.js
Dwaynekj: Also. We should have warnings if a plugin collides with an existing property on one of the Core objects. As in someone wrote a method and mistakenly called it setData

Dwaynekj:
@rdepena I think this one is important for you to look at as well.
@joneit has a plugins implementation, but we don't really have a product definition on what a Hypergrid plugin `is` and `isn't`. Or is it anything?

rdepena:
@Dwaynekj @joneit I kinda agree with DJ, can we take a step back instead of reviewing the code, can we have a high level conversation on what plugins are and how they integrate into hypergrid.

Dwaynekj:
@rdepena @joneit Discussed in person. We will expose this architecture with the caveat that its exploratory and it is subject to changed. We must the user well aware of this.

I do feel it may be too powerful and would be hard to rescind the idea later. If we had a formal and properly pruned and public API (long time coming now) then we would expose hooks that plugins would be allowed to use.

At the moment we are saying a plugin can do anything it wants.

👍

joneit:
* Changed name of plugins hash from `grid.plugin` to `grid.plugins` _**throughout**_
* Added `unintallPlugins` method, which can be called like this:
```javascript
grid.uninstallPlugins(); //uninstalls all
grid.uninstallPlugins('myPlugin'); //uninstalls just this one
grid.uninstallPlugins(['myPlugin', 'myOtherPlugin']);
```

Dwaynekj:
🎉

PR #342, opened 2016-09-22
IMPR/Sorting-API

Dwaynekj:
@joneit @rdepena Paired PR with https://github.com/openfin/hyper-analytics/pull/34

### DIFF
--- a/add-ons/hyper-sorter/mix-ins/dataModel.js
+++ b/add-ons/hyper-sorter/mix-ins/dataModel.js
@@ -1,51 +1,67 @@
 'use strict';
 
+var UPWARDS_BLACK_ARROW = '\u25b2', // aka '▲'
+    DOWNWARDS_BLACK_ARROW = '\u25bc'; // aka '▼'
+
 module.exports = {
 
     /**
      * @memberOf dataModels.JSON.prototype
-     * @param {number} colIndex
+     * @param column
      * @param keys
      */
-    toggleSort: function(colIndex, keys) {
-        this.incrementSortState(colIndex, keys);
-        this.applyAnalytics({columnSort: true});
+    toggleSort: function(column, keys) {
+        this.incrementSortState(column, keys);
+        this.serializeSortState();
+        this.reindex();
     },
     /**
      * @memberOf dataModels.JSON.prototype
-     * @param {number} columnIndex
+     * @param column
      * @param {boolean} deferred
      */
-    unSortColumn: function(columnIndex, deferred) {
+    unSortColumn: function(column, deferred) {
         var sorts = this.getSortedColumnIndexes(),
-            sortPosition;
+            result = this.getColumnSortState(column.index),
+            sortPosition = result.sortPosition;
 
-        if (sorts.find(function(sortSpec, index) {
-                sortPosition = index;
-                return sortSpec.columnIndex === columnIndex;
-            })) {
-            sorts.splice(sortPosition, 1);
+        if (sortPosition > -1) {
+            sorts.splice(sortPosition, 1); //Removed from sorts
             if (!deferred) {
-                this.applyAnalytics({columnSort: true});
+                this.sorter.prop('columnSorts', sorts);
+                this.reindex();
             }
         }
+        this.serializeSortState();
     },
-    /**
-     * @memberOf dataModels.JSON.prototype
-     * @param {number} colIndex
-     * @param {string[]} keys
-     */
-    incrementSortState: function(columnIndex, keys) {
+
+    getColumnSortState: function(columnIndex){
         var sorts = this.getSortedColumnIndexes(),
+            sortPosition = -1,
             sortSpec = sorts.find(function(spec, index) {
+                sortPosition = index;
                 return spec.columnIndex === columnIndex;
             });
+        return {sortSpec: sortSpec, sortPosition: sortPosition};
+    },
+
+    /**
+     * @memberOf dataModels.JSON.prototype
+     * @param column
+     * @param {string[]} keys
+     * @return {object[]} sorts
+     */
+    incrementSortState: function(column, keys) {
+        var sorts = this.getSortedColumnIndexes(),
+            columnIndex = column.index,
+            sortSpec = this.getColumnSortState(columnIndex).sortSpec;
 
         if (!sortSpec) { // was unsorted
             if (keys.indexOf('CTRL') < 0) { sorts.length = 0; }
             sorts.unshift({
                 columnIndex: columnIndex, // so define and...
-                direction: 1 // ...make ascending
+                direction: 1, // ...make ascending
+                type: column.getType()
             });
         } else if (sortSpec.direction > 0) { // was ascending
             sortSpec.direction = -1; // so make descending
@@ -57,5 +73,40 @@ module.exports = {
         if (sorts.length > 3) {
             sorts.length = 3;
         }
+    },
+
+    serializeSortState: function(){
+        var state = this.getPrivateState();
+        state.sorts = this.getSortedColumnIndexes();
+    },
+
+    /**
+     * @memberOf dataModels.JSON.prototype
+     * @desc returns the columns that currently sorted and their intended direction of the sort
+     */
+    getSortedColumnIndexes: function() {
+        return this.sorter.prop('sorts') || [];
+    },
+    /**
+     * @memberOf dataModels.JSON.prototype
+     * @param index
+     * @param returnAsString
+     * @desc Provides the unicode character used to denote visually if a column is a sorted state
+     * @returns {*}
+     */
+    getSortImageForColumn: function(columnIndex) {
+        var sorts = this.getSortedColumnIndexes(),
+            state = this.getColumnSortState(columnIndex),
+            sortSpec = state.sortSpec,
+            sortPosition = state.sortPosition,
+            result = null;
+
+        if (sortSpec) {
+            var rank = sorts.length - sortPosition,
+                arrow = sortSpec.direction > 0 ? UPWARDS_BLACK_ARROW : DOWNWARDS_BLACK_ARROW;
+            result = rank + arrow + ' ';
+        }
+
+        return result;
     }
 };
--- a/demo/js/aggregations-view.js
+++ b/demo/js/aggregations-view.js
@@ -28,12 +28,13 @@ window.onload = function() {
         plugins: [
             Hypergrid.drillDown, // simple API install (plain object with `install` method) but no `name` defined so no ref is saved
             Hypergrid.Hyperfilter, // object API instantiation; `$$CLASS_NAME` defined so ref saved in `grid.plugins.hyperfilter`
-            Hypergrid.Hypersorter, // object API instantiation to grid.plugins; no `name` or `$$CLASS_NAME` defined so no ref saved
+            [Hypergrid.Hypersorter, {Column: fin.Hypergrid.behaviors.Column}], // object API instantiation to grid.plugins; no `name` or `$$CLASS_NAME` defined so no ref saved
             [Hypergrid.AggregationsView, options] // object API instantiation with one arg; `$$CLASS_NAME` defined so ref saved in `grid.plugins.aggregationsView`
         ]
     });
 
     grid.filter = grid.plugins.hyperfilter.create();
+    grid.sorter = grid.plugins.hypersorter;
 
     // show filter row as per `options`
     grid.setState({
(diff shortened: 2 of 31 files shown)


demo/simple-column-sorting-pipeline.html
```diff
@@ -0,0 +1,155 @@
+<!DOCTYPE html>
```
Dwaynekj: @rdepena As an outsider, can you comment if this makes sense. You only need to review this file.

Dwaynekj: It does work

joneit: so this is nice but, as I've mentioned, I believe the purpose of the ticket is to **provide a way to replace the comparators** — not necessarily the the entire sorting transformer.

_(I do understand the advantage you explained to me, that it no longer uses Steve Wirt's code. Not sure why that's a priority. Steve did some interesting things and it has a number of advantages, specifically that it handles multi-column sorts and does so relatively efficiently.)_

So to complete this ticket I believe we need to provide a way to specify your own comparators. I already gave you my thoughts on that:

1. When a comparator is given for a column, use it!
2. When a comparator is not given, but a type is, use that to pick a standard comparator.
3. When type is not known, infer number vs string from first row.

Item 3 above is the way it works now. Let's add 1 and 2.

I also pointed out the following observations about the way it chooses comparators now:

1. It uses separate comparators for number (a-b) and string (a < b, a > b). The latter would clearly work for both but I suppose the thinking was that the former is more efficient for numbers (one arithmetic operation instead of two)? Maybe we should profile or jsPerf it to see if it make a significant difference.
2. It uses separate ascending and descending comparators. This is necessary because Array#sort does not have an option for direction. So our app developers will need to supply comparators in pairs.

I also suggested that comparators be supplied in the same way we are currently supplying `fields` (column names) and `calculators`, which are discrete array parameters to DataSourceOrigin constructor. We could just add another parameter:


```javascript
new DataSourceOrigin(data, fields, calculators, comparators)
```

**_HOWEVER_** on further reflection, I think it is time we supplied a single array instead:

```javascript
new DataSourceOrigin(data, fields)
```
where _each element_ of `fields` would be defined as follows:
* string - column name
* object with following properties:
   * `name` _(required)_ column name
   * `calculator:` _(optional)_ a function for a computed column (only)
   * `comparator:` _(optional)_ object:
      * `asc:` ascending comparator
      * `desc:` descending comparator

Dwaynekj: I admittedly, pivoted away from our original plans, but I feel this is the approach to take.

With regards to keeping the existing DataSourceSorter paradigm..

I think the point is not to get rid of Steve Wirt's code but decide who is owning what portion (us or the application developer). The code now clearly puts the onus on us to maintain all uses cases for sorting all Hypergrid users could potentially want.  The scaffold for sorting has opinions in that it assumes the user wants a straightforward multi-column sort for instance. What if they want only single column sort? What if they want to not sort certain columns against certain other columns? What if they want they only want to have certain cells in a column fixed and unsorted? I think having this scaffold to support simply comparators would not suffice and we need to emphasize that building your own sorting pipeline for in-memory is easy to do.

I do agree that DataSourceSorter has nice patterns, but I can't anticipate being generic enough. The idea is to empower the user and not have them dependent on our patterns.

With regards to simply having a default comparator option...
I don't think any defaults should be created. Maybe always alphanumeric sorting will be standard sorting, but what happens for different data structures (especially since our scaffold wants to be able to do multi-column sort) what do we assume the comparator is for a sparkline or a button even? This puts the maintenance of this code on us where as if we just make it easy for the user to make their own sorter then they can control it fully.

demo/js/aggregations-view.js
```diff
@@ -28,16 +28,17 @@ window.onload = function() {
...
         ]
     });
 
     grid.filter = grid.plugins.hyperfilter.create();
+    grid.sorter = grid.plugins.hypersorter;
 
     // show filter row as per `options`
     grid.setState({
-        showFilterRow: options.includeFilter && grid.filterProp('columnFilters')
+        showFilterRow: options.includeFilter && grid.dataSourceProp('filter', 'columnFilters')
```
joneit: Let's go with shorter/sweeter `dataProp` if you don't mind

src/dataModels/JSON.js
```diff
@@ -748,39 +762,40 @@ var JSON = DataModel.extend('dataModels.JSON', {
      *
      * @memberOf dataModels.JSON.prototype
      */
-    filterProp: function(columnIndex, property, value) {
-        var properties = filterPropsFromParams.apply(this, arguments),
-            result = this.filter.prop(properties);
+    dataSourceProp: function(dataSource, columnIndex, property, value) {
+        var properties = propsFromParams.apply(this, Array.prototype.slice.call(arguments, 1, arguments.length)),
```
joneit: You don't need the last param `arguments.length` here

joneit:
In Columns.js, comment off these lines so all types of numbers are `'number'`:
```javascript
            if (type !== eachType) {
                if (isNumber && (typeof value === 'number')) {
                    type = 'float';
                } else {
                    return 'mixed';
                }
            }
```

demo/js/aggregations-view.js
Dwaynekj: This has been revised as we discussed. We will clarify this tomorrow

Dwaynekj:
Will also try to get Hyperanalytics to run with Hypergrid's copy of Base

demo/js/aggregations-view.js
Dwaynekj: Done on JE PR

src/behaviors/Behavior.js
```diff
@@ -197,8 +197,15 @@ var Behavior = Base.extend('Behavior', {
         console.error('getCellProvider() is deprecated as of v1.0.6. No replacement; do not call. Previously called by `Behavior` constructor; `new CellRenderers()` is now called by `Hypergrid` constructor instead.', arguments);
     },
 
+    /**
+     * @deprecated
+     * @memberOf Behavior.prototype
+     */
     applyAnalytics: function() {
-        this.dataModel.applyAnalytics();
+        return this.deprecated('applyAnalytics()', 'reIndex()', '1.1.0', arguments);
```
Dwaynekj: N/A

src/behaviors/Behavior.js
```diff
@@ -197,8 +197,15 @@ var Behavior = Base.extend('Behavior', {
...
+    },
+    reIndex: function() {
```
Dwaynekj: DONE

src/dataModels/JSON.js
```diff
@@ -748,39 +762,40 @@ var JSON = DataModel.extend('dataModels.JSON', {
      *
      * @memberOf dataModels.JSON.prototype
      */
-    filterProp: function(columnIndex, property, value) {
-        var properties = filterPropsFromParams.apply(this, arguments),
-            result = this.filter.prop(properties);
+    dataSourceProp: function(dataSource, columnIndex, property, value) {
```
Dwaynekj: N/A. Fixed in JE PR

src/dataModels/JSON.js
```diff
@@ -748,39 +762,40 @@ var JSON = DataModel.extend('dataModels.JSON', {
...
+        var properties = propsFromParams.apply(this, Array.prototype.slice.call(arguments, 1, arguments.length)),
+            ds = this[dataSource],
```
Dwaynekj: N/A. Fixed in JE PR

src/Hypergrid.js
```diff
@@ -95,10 +97,10 @@ function Hypergrid(container, options) {
     container = container || findOrCreateContainer(options.boundingRect);
     this.setContainer(container);
 
-    if (options.Behavior) {
-        this.setBehavior(options); // also sets options.pipeline and options.data
-    } else if (options.data) {
-        this.setData(options.data); // if no behavior has yet been set, also sets default behavior and options.pipeline
+    if (this.options.Behavior) {
```
Dwaynekj: Just as a guard against a future change.

src/dataModels/JSON.js
```diff
@@ -837,7 +852,7 @@ function reselectGridRowsBackedBySelectedDataRows() {
     if (this.selectedData.length) { // any data row objects added from previous grid row selections?
         var selectionModel = this.grid.selectionModel,
             offset = this.grid.getHeaderRowCount(),
-            filteredData = this.getFilteredData();
+            filteredData = this.getReIndexedData();
```
Dwaynekj: Done

src/dataModels/JSON.js
```diff
@@ -890,8 +905,10 @@ function filterPropsFromParams(columnIndex, propName, value) {
     if (isColumnProp) {
         // non-enumerable propName:
         Object.defineProperty(properties, 'column', {
-            index: columnIndex,
-            name: this.source.getFields()[columnIndex]
+            value: {
```
Dwaynekj: DONE

src/dataSources/DataSourceBase.js
```diff
@@ -112,6 +116,14 @@ DataSourceBase.prototype = {
             return this.dataSource.isDrillDown.apply(this.dataSource, arguments);
         }
     },
+    getReIndexedData: function() {
```
Dwaynekj: DONE

src/dataSources/DataSourceBase.js
```diff
@@ -168,7 +180,28 @@ DataSourceBase.prototype = {
             data[r] = row;
         }
         console.table(data);
-    }
+    },
+
+    transform: capitalize,
```
Dwaynekj: DONE

src/dataSources/DataSourceBase.js
```diff
@@ -168,7 +180,28 @@ DataSourceBase.prototype = {
...
 };
 
+function capitalize(string) {
```
Dwaynekj: DONE

src/dataSources/DataSourceBase.js
```diff
@@ -168,7 +180,28 @@ DataSourceBase.prototype = {
...
+
+function replacer(a, b, c) {
+    return b.toUpperCase() + c;
+}
+
+// function set(methodName) {
+//     capitalize.transform = capitalize[methodName];
+// }
+//
+// function passthrough(string) {
```
Dwaynekj: DONE

src/index.js
```diff
@@ -22,6 +22,7 @@ Hypergrid.images = require('../images');
 Hypergrid.behaviors = require('./behaviors');
 Hypergrid.dataModels = require('./dataModels');
 Hypergrid.features = require('./features');
+Hypergrid.pipelineBase = require('./dataSources/DataSourceBase');
```
Dwaynekj: DONE

src/dataModels/JSON.js
Dwaynekj: N/A. DONE on JE PR

src/dataSources/DataSourceOrigin.js
```diff
@@ -195,7 +198,7 @@ var DataSourceOrigin = DataSourceBase.extend('DataSourceOrigin',  {
              * @memberOf DataSource#
              */
             this.headers = this.headers || this.getDefaultHeaders().map(function(each) {
-                return headerify.transform(each);
+                return self.transform(each);
```
Dwaynekj: DONE

add-ons/hyper-sorter/mix-ins/dataModel.js
```diff
@@ -1,51 +1,67 @@
...
+        this.reIndex();
     },
     /**
      * @memberOf dataModels.JSON.prototype
-     * @param {number} columnIndex
+     * @param col
      * @param {boolean} deferred
      */
-    unSortColumn: function(columnIndex, deferred) {
+    unSortColumn: function(col, deferred) {
```
Dwaynekj: DONE

demo/simple-column-sorting-pipeline.html
```diff
@@ -0,0 +1,158 @@
...
+            },
+            // This function is called on every reIndex call if this object is in the pipelne
+            apply: function (){
+                var c = this.sorter.columnIndex;
+                if (c === null) {
+                    this.data.length = 0;
+                    return;
+                }
+
+                var header = this.dataSource.getFields()[c];
```
Dwaynekj: FIXED

src/dataSources/DataSourceOrigin.js
```diff
@@ -227,6 +230,10 @@ var DataSourceOrigin = DataSourceBase.extend('DataSourceOrigin',  {
         this.headers = headers;
     },
 
+    getReIndexedData: function() {
```
Dwaynekj: DONE

PR #357, opened 2016-10-12
container creation, naming, css class

joneit:
Removed default container size/position styles from code and put in new css class `hypergrid-container`. There was a problem with instantiating multiple grids on the same page when depending on the container creation code. That code first looked for a container id="hypergrid" and if found, use it; if not found create it. So first time it created it; subsequently it found it, so kept appending additional grids inside it. I fixed this by
1. requring the found container to be empty in order to use it. 
2. Moved the naming to the initContainer function, but instead of always naming it hypergrid, I named it hypergrid, hypergrid1, hypergrid2, hypergrid3, etc.

### DIFF
--- a/css/grid.css
+++ b/css/grid.css
@@ -1,3 +1,7 @@
+.hypergrid-container {
+    position: relative;
+    height: 500px;
+}
 .hypergrid-textfield {
     position: absolute;
     font-size: 12px;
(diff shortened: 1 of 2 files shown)


Dwaynekj:
We can talk about this one in person tomorrow

PR #363, opened 2016-10-18
Bug/scroll on focus

noahtaylor:
I have experienced an issue whereby the screen can scroll after an editor shown.  I did some debugging and found the issue occurs because of the focus change that occurs on the editor element.  Further this was found as a resolution:  http://stackoverflow.com/questions/4898203/jquery-focus-without-scroll

I added that change to prevent window from scrolling after a cell editor has taken focus.  I tested the fix in chrome and firefox but was unable to test in safari and ie 10+ as I do not have access to these browsers.

### DIFF
--- a/src/cellEditors/CellEditor.js
+++ b/src/cellEditors/CellEditor.js
@@ -393,7 +393,9 @@ var CellEditor = Base.extend('CellEditor', {
 
         el.style.left = el.style.top = 0; // work-around: move to upper left
 
+        var x = window.scrollX, y = window.scrollY;
         this.input.focus();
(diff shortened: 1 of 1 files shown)


Dwaynekj:
I changed it to base to develop

Dwaynekj:
@noahtaylor  Can you help us reproduce?

noahtaylor:
check this out.  https://jsfiddle.net/g51f5628/11/
scroll to bottom then click on cells from the top grid to make it scroll up.

Dwaynekj:
Ok thanks. We will review for our Monday release

PR #368, opened 2016-10-25
IMPRV/DataSourceOrigin:FindRowByIndex

Dwaynekj:
Dependent on merging https://github.com/openfin/fin-hypergrid/pull/367

### DIFF
--- a/src/dataSources/DataSourceOrigin.js
+++ b/src/dataSources/DataSourceOrigin.js
@@ -177,6 +177,7 @@ var DataSourceOrigin = DataSourceBase.extend('DataSourceOrigin',  {
      * * _omitted_ - Ignored.
      * * _object_ - Replacement for the data row if found.
      * * `null` - Flag to delete the data row if found. The found data row is nonetheless returned.
+     * * `undefined` - Flag to delete the row at that index.
      * @returns {object|number|undefined} One of:
      * * `undefined` - data row not found
      * * _object_ - found data row object (will have been deleted if `replacement` was `null`)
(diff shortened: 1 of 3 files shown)


src/dataSources/DataSourceOrigin.js
```diff
@@ -170,6 +170,47 @@ var DataSourceOrigin = DataSourceBase.extend('DataSourceOrigin',  {
...
+        var result;
+
+        if (arguments.length < 1) {
+            throw 'Expected at least 1 argument but found ' + arguments.length + '.';
+        }
+
+        if (typeof index !== 'number') {
+            throw 'Expected at index to be a number but got ' + index + '.';
+        }
+        result = this.data.find(function(row, idx) {
```
joneit: No! Use `getDataIndex` here. Explanation: We need to call getDataIndex to transpose all the indexes. In other words, if the filtered & sorted dataset has 10 rows left out of 1000 and we want to delete row the 7th row, calling `getDataIndex(7-1)` will return the underlying index of that row in `source.data`. Get it?

src/dataSources/DataSourceOrigin.js
```diff
@@ -170,6 +170,47 @@ var DataSourceOrigin = DataSourceBase.extend('DataSourceOrigin',  {
...
+            if (!row) { return; }
+            return index === idx;
+        });
+
+        if (result) {
+            if (replacement === null) {
+                this.data.splice(index, 1);
+            } else if (typeof replacement === 'object') {
+                this.data[index] = replacement;
+            } else if (replacement !== undefined) {
```
joneit: Replace this line with:
```javascript
            } else if (replacement === undefined && arguments.length >= 2) {
                delete this.data[index];
            } else {
```
Explanation: My comment was wrong. `undefined` should delete the row element. This leaves `undefined` in the element which renders as a blank row.

src/dataSources/DataSourceOrigin.js
Dwaynekj: I think we should encourage the application developer to implement what your describing. However at DataSourceOrigin base we should allow the ability to delete the underlying data. Thoughts?

src/dataSources/DataSourceOrigin.js
Dwaynekj: Done

src/dataSources/DataSourceOrigin.js
joneit: as discussed we'll leave calling `getDataIndex` to the caller to pre-translate his index if that's what he wants to do... **but hey DJ you don't need `.find(...)` here! just do `this.data[index]`, right?**

joneit: wow I have a better idea. if we put `getDataIndex` in the function, then people can still call it both ways:
```
dataModel.source.findRowByIndex(n, null) // deletes actual row n
dataModel.dataSource.findRowByIndex(n, null) // deletes transformed (displayed) row n
```

Dwaynekj: oh duh! Good Catch

PR #376, opened 2016-11-01
v1.2.2 patch: getCell config should not be sticky

joneit:
@Dwaynekj One of the optimizations I put in 1.2.1 makes `getCell` changes to `config` look-and-feel props sticky till the bottom of each column. This was unintended. This patch restores it and I believe we should push this soon (like today). Unfortunately, we loose the ~1 FPS that was gained by this change.

### DIFF
--- a/src/behaviors/Behavior.js
+++ b/src/behaviors/Behavior.js
@@ -289,14 +289,10 @@ var Behavior = Base.extend('Behavior', {
      * @return {object} Newly created default empty tablestate.
      */
     getDefaultState: function() {
-        var tableProperties = this.grid._getProperties();
-        var state = Object.create(tableProperties);
-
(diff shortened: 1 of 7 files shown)


src/behaviors/Behavior.js
```diff
@@ -289,14 +289,10 @@ var Behavior = Base.extend('Behavior', {
...
-        var state = Object.create(tableProperties);
-
-        _(state).extendOwn({
-            columnProperties: []
-        });
-
-        return state;
+        return Object.create(
+            this.grid._getProperties(),
+            { columnProperties: { value: [] } }
```
Dwaynekj: @joneit  columnProperties is not initialized on grid.properties already?

PR #377, opened 2016-11-01
draw rect + v-lines per column + h-lines after

joneit:
Instead of doing this (v1.2.1):
* Set CSS backgroundColor style of canvas element to "line color."
* Draw a separate little rect for each cell.

We now do this (v1.2.2, proposed):
* Draw a single rect for the column.
* Draw a vertical line to left of each column rect.
* Draw a cell rect when and only when the background color of the cell differs from the column.
* After all columns are rendered, draw a horizontal line below each visible row.

Also retired feature where `config.backgroundColor`, `config.backgroundSelectionColor`, and `config.foregroundSelectionColor` can all be functions instead of strings. These are the only such properties that allow this. I suspect this feature preceded `getCell` because we never use it and it could also be done in`getCell`.

### DIFF
--- a/demo/js/demo.js
+++ b/demo/js/demo.js
@@ -459,7 +459,7 @@ window.onload = function() {
                         case 5:
                             config.backgroundColor = '#e8ffe8';
                             config.font = 'italic x-small verdana';
-                            if (config.color !== redIfStartsWithS) {
+                            if (x !== idx.LAST_NAME) {
                                 config.color = '#070';
                             }
                             break;
@@ -475,6 +475,8 @@ window.onload = function() {
 
                 switch (x) {
                     case idx.LAST_NAME:
+                        config.color = config.value != null && (config.value + '')[0] === 'S' ? 'red' : '#191919';
+                        // eslint-disable-line no-fallthrough
                     case idx.FIRST_NAME:
                     case idx.BIRTH_STATE:
                     case idx.RESIDENCE_STATE:
@@ -1054,7 +1056,6 @@ window.onload = function() {
         //                    });
 
         behavior.setColumnProperties(idx.LAST_NAME, {
-            color: redIfStartsWithS,
             columnHeaderBackgroundColor: '#142B6F', //dark blue
             columnHeaderColor: 'white'
         });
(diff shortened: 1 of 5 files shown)


demo/js/demo.js
```diff
@@ -475,6 +475,8 @@ window.onload = function() {
 
                 switch (x) {
                     case idx.LAST_NAME:
+                        config.color = config.value != null && (config.value + '')[0] === 'S' ? 'red' : '#191919';
```
Dwaynekj: I know its the demo, but lets use !==

src/lib/Renderer.js
```diff
@@ -989,6 +1019,7 @@ var Renderer = Base.extend('Renderer', {
         config.isRowSelected = isRowSelected;
         config.isColumnSelected = isColumnSelected;
         config.isInCurrentSelectionRectangle = grid.isInCurrentSelectionRectangle(x, r);
+        config.columnBackgroundColor = columnProperties.backgroundColor; // could be function?
```
Dwaynekj: Remove that comment

demo/js/demo.js
joneit: `value == null` and `value != null` are special cases acceptable to the linter meaning `value === undefined || value === null` and `value !== undefined && value !== null`, respectively.(This was existing code; I just moved it.)

src/lib/Renderer.js
joneit: Right!

src/lib/Renderer.js
```diff
@@ -893,6 +905,24 @@ var Renderer = Base.extend('Renderer', {
...
+     */
+    paintGridlines: function(gc) {
+        if (this.grid.properties.gridLinesH) {
+            var viewWidth = this.visibleColumns[this.visibleColumns.length - 1].right,
+                lineWidth = this.grid.properties.lineWidth;
+
+            gc.fillStyle = this.grid.properties.lineColor;
+
+            this.visibleRows.forEach(function(visibleRow) {
+                gc.fillRect(0, visibleRow.bottom, viewWidth, lineWidth);
```
joneit: Yes, both horizontals and verticals.

PR #380, opened 2016-11-02
Omit GraphicsContext object when no options.logger

joneit:
Now uses grid.properties.doubleClickDelay, which if < 100ms turns off double click detection altogether.

### DIFF
--- a/src/lib/Canvas.js
+++ b/src/lib/Canvas.js
@@ -1,0 +1,762 @@
+/* eslint-env browser */
+
+'use strict';
+
+if (typeof window.CustomEvent !== 'function') {
+    window.CustomEvent = function(event, params) {
+        params = params || { bubbles: false, cancelable: false, detail: undefined };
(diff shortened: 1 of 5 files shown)


src/lib/Canvas.js
```diff
@@ -0,0 +1,762 @@
...
+        }
+        e.preventDefault();
+        this.dispatchNewMouseKeysEvent(e, 'fin-canvas-wheelmoved', {
+            isRightClick: this.isRightClick(e)
+        });
+    },
+
+    finclick: function(e) {
+        var delay = this.component.resolveProperty('doubleClickDelay');
+        if (delay < 100) {
```
Dwaynekj: We're overriding the user if they go under 100?

joneit: Yes, 100ms is the low end of the range for double clicking. Anything less than that is ridiculous.

Dwaynekj: What was it b4

PR #384, opened 2016-11-05
GRID-414: Cache <canvas>'s graphics context properties

joneit:
Hypergrid speed improvement of 24.4% (Chrome 54) and 19.4% (Chrome 40) on Mac OS.

See [the Jira ticket](https://appoji.jira.com/browse/GRID-414) for details.


src/Hypergrid.js
```diff
@@ -1564,7 +1564,7 @@ var Hypergrid = Base.extend('Hypergrid', {
         this.resizeScrollbars();
     },
     synchronizeScrollingBoundries: function() {
-        this.deprecated('synchronizeScrollingBoundries', 'synchronizeScrollingBoundaries', '1.2.0');
+        this.deprecated('synchronizeScrollingBoundries()', 'synchronizeScrollingBoundaries()', '1.2.0');
```
Dwaynekj: 👍

src/lib/GraphicsContext.js
```diff
@@ -1,72 +0,0 @@
-'use strict';
```
Dwaynekj: This file was solely for debugging

src/lib/Canvas.js
```diff
@@ -759,4 +740,50 @@ function makeCharMap() {
...
+
+    function makeStub(key) {
+        if (
+            !(key in props) &&
+            !/^(webkit|moz|ms|o)[A-Z]/.test(key) &&
+            typeof gc[key] !== 'function'
+        ) {
+            Object.defineProperty(props, key, {
+                get: function() {
+                    return (values[key] = values[key] || gc[key]);
```
Dwaynekj: Ok if understand correctly you just putting a facade infront of any property access to the gc obj

Dwaynekj: Are we sure there is no native functionality offering this feature. It seems like a common enough thing to avoid RW

joneit: It does seem odd. But I haven't been able to find any reference to any such thing. I did find jsperfs showing the dramatic difference between doing this and not doing it.

src/lib/GraphicsContext.js
joneit: Yes. If we need in future we can resurrect. I did not want to spend time on making it right now.

src/lib/Renderer.js
```diff
@@ -1067,23 +1066,6 @@ var Renderer = Base.extend('Renderer', {
         return this.buttonCells[key] === true;
     },
 
-    startAnimator: function() {
```
joneit: Apparently not. There was that comment `// animate the dashed line a bit here for fun` that appeared in several places. Apparently steve did this "for fun" at one point but then stopped using it. Looks like it made a moving "marquee."

src/lib/Canvas.js
joneit: BTW: The mediator is not mandatory — but of course the cache would get out of sync on setter calls and getter calls would be way, way slower.

Dwaynekj: Ok

PR #385, opened 2016-11-06
GRID-415: Measure text by individual characters

joneit:
**Requires GRID-414 (https://github.com/openfin/fin-hypergrid/pull/384) to be merged first before running.**

### DIFF
--- a/src/defaults.js
+++ b/src/defaults.js
@@ -2,8 +2,6 @@
 
 'use strict';
 
-var LRUCache = require('lru-cache');
-
 var ALPHA_REGEX = /^(transparent|((RGB|HSL)A\(.*,\s*([\d\.]+)\)))$/i;
 
 /**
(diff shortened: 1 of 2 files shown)


src/defaults.js
```diff
@@ -834,7 +832,7 @@ var defaults = {
  * @see https://developer.mozilla.org/en-US/docs/Web/CSS/font
  */
 
-var textWidthCache = new LRUCache(2000);
+var fontMetrics = {};
```
Dwaynekj: I don't see why the cache needed to be removed, but its not like we expect a million differents fonts on one grid

joneit: The LRU cache cached the most recent 2000 strings (cell values) and their pixel widths. We're not doing that any more. The new measure is so much faster it's probably not worth it. Only really helped on a static screen, since scrolling would wipe those 2000 in short order. But on a static screen, I believe our partial render design will obviate the need for this. See what I mean?

PR #389, opened 2016-11-08
GRID-420: Row properties

joneit:


demo/js/demo.js
```diff
@@ -452,25 +452,6 @@ window.onload = function() {
                     n = behavior.getColumn(idx.TOTAL_NUMBER_OF_PETS_OWNED).getValue(y);
                     hex = (155 + 10 * (n % 11)).toString(16);
                     config.backgroundColor = '#' + hex + hex + hex;
-                } else {
-                    switch (config.y % 6) {
```
Dwaynekj: While we're at it. We can simplify getCell even further. Removing https://github.com/openfin/fin-hypergrid/pull/389/files#diff-c4ec3078804b5269f20ac55a6f14d676R443

demo/js/demo.js
```diff
@@ -982,6 +963,16 @@ window.onload = function() {
                 // idx.SQUARE_OF_INCOME
             ],
 
+            backgroundColor: 'white',
+            gridRowProperties: [
```
Dwaynekj: I think `rowProperties` is good enough

demo/js/demo.js
```diff
@@ -982,6 +963,16 @@ window.onload = function() {
...
+                undefined,
```
Dwaynekj: I think it would be cleaner to have object with the row index as keys as opposed to a sparse array

src/lib/Renderer.js
```diff
@@ -951,6 +950,8 @@ var Renderer = Base.extend('Renderer', {
             isSelected = isRowSelected || grid.isCellSelectedInRow(r);
         } else if (isGridRow) {
             config = Object.create(cellEvent.properties);
+            alternates = cellEvent.properties.gridRowProperties;
```
Dwaynekj: What is this line doing?

demo/js/demo.js
joneit: Actually, I named it that because this is not a general implementation. It is _restricted_ to data rows ("grid rows") and is ignored for other subgrids. But your suggestion _is_ in keeping with the way properties are named in general (`backgroundColor` for grid rows, `columnHeaderBackgroundColor` for a column's header cell). **So I will make the change.**

demo/js/demo.js
joneit: I tightened it up a bit.

demo/js/demo.js
joneit: Read the comments and the ticket. This is a array that represents a repeating pattern. I'll try to make it clearer in the comments.

src/lib/Renderer.js
joneit: At first I called the property `alternateRowProperties` because it was just for alternating row property bags. I'll rename the local var `rowProperties`.

PR #402, opened 2016-11-21
GRID-430: High-performant grid renderers - additional refinements

joneit:


Dwaynekj:
Well most of the core logic was in the other PR

PR #404, opened 2016-11-25
Fix selection region frame

joneit:


Dwaynekj:
@joneit Whats this about? Can you lmk b4 you merge them, so I can understand what your doing?

PR #405, opened 2016-11-26
GRID-373: Grid state, continued (part 2)

joneit:
The subgrids option is now serializable. That is, you can now give strings instead of a references to data models. The strings are registered data model names (in src/dataModels/index.js).

Like grid renderers, these data model names are case-sensitive. This is however a break from our other name registries, such as cell renderers and cell editors. We clearly should be consistent, but rather than following suit, I'm thinking of making cell renderers and cell editors case-sensitive as well because its simpler, less code, more performant (cell renderers require a .toLowerCase() call on every cell on every render), and not too much to ask from developers who are after all already familiar with case-senstive identifers in Javascript. The only reason SWirts made them case insensitive was that as data in a datagram it seemed safer. Thoughts?

This is part of GRID-373, our drive towards putting all critical set-up data in a serializable state datagram that can be used in a setState call, or passed to a Hypergrid constructor.

### DIFF
--- a/add-ons/drill-down.js
+++ b/add-ons/drill-down.js
@@ -5,7 +5,7 @@
 /**
  * Methods for programmatic drill-down manipulation.
  *
- * Intended to be mixed into Hypergrid's in-memory data model object (./src/dataModels/JSON.js) using the included {@link drillDown.mixInTo}
+ * Intended to be mixed into Hypergrid's in-memory data model object ({@link dataModels.JSON}) using the included {@link drillDown.mixInTo}
  * _Do not use directly._
  *
  * For example, you can mix it into the data model prototype like this:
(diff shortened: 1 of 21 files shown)


Dwaynekj:
It seems some of this pr was in the other branch that you merged? Can you help me understand this half of it? Or is it just cleanup?

joneit:
DJ, in answer to your question (I had hoped the PR description was clear, but to reiterate): This was just adding subgrid _name_ as an option. Previously, it accepted only _a reference to_ a subgrid. I'm talking about the Behavior subgrids setter, which accepts a list of subgrids. By "registered" I mean that it is included in the hash of known data models (in src/dataModels/index.js).

As an aside, we do need to stop pre-registering everything, which forces them to be part of the build. By including everything, the build is much bigger than it needs to be. We need to begin letting developers pick and choose what to include (either through `<script>` tags or by using Browserify).

Dwaynekj:
Ah ok. What is the default if no option is passed?

PR #412, opened 2016-12-05
GRID-431: Misaligned filter cell editor

joneit:
See Jira [GRID-431](https://appoji.jira.com/browse/GRID-431) for details.

### DIFF
--- a/src/Hypergrid.js
+++ b/src/Hypergrid.js
@@ -618,11 +618,9 @@ var Hypergrid = Base.extend('Hypergrid', {
      * @memberOf Hypergrid#
      */
     getMouseDown: function() {
-        var last = this.mouseDown.length - 1;
-        if (last < 0) {
-            return null;
+        if (this.mouseDown.length) {
+            return this.mouseDown[this.mouseDown.length - 1];
         }
-        return this.mouseDown[last];
     },
 
     /**
@@ -630,7 +628,7 @@ var Hypergrid = Base.extend('Hypergrid', {
      * @desc Remove the last item from the mouse down stack.
      */
     popMouseDown: function() {
-        if (this.mouseDown.length !== 0) {
+        if (this.mouseDown.length) {
             this.mouseDown.length = this.mouseDown.length - 1;
         }
     },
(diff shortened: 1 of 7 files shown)


src/Hypergrid.js
```diff
@@ -1075,7 +1073,7 @@ var Hypergrid = Base.extend('Hypergrid', {
             event.isDataColumn &&
             event.getCellProperty(event.isDataRow ? 'editable' : 'filterable')
         ) {
-            this.setMouseDown(event.gridCell);
+            this.setMouseDown(new Point(event.gridCell.x, event.dataCell.y));
```
Dwaynekj: they take Points now?

src/cellEditors/CellEditor.js
```diff
@@ -242,8 +233,6 @@ var CellEditor = Base.extend('CellEditor', {
             this.grid.cellEditor = null;
             this.el.remove();
         } else if (feedback >= 0) { // never true when `feedback` undefined
-            var point = this.event.gridCell;
-            this.grid.selectViewportCell(point.x, point.y - this.grid.getHeaderRowCount());
```
Dwaynekj: This removes the selection?

src/defaults.js
```diff
@@ -653,6 +653,17 @@ var defaults = {
...
     /**
+     * This is used only by FilterBox cell editor.
+     * One of:
+     * * **`'onCommit'`** - Column filter state not set until keyup === `\r` (return/enter key)
+     * * **`'immediate'`** - Column filter state set on each key press
+     * @default
+     * @type {boolean}
+     * @memberOf module:defaults
+     */
+    filteringMode: 'onCommit',
```
Dwaynekj: My instinctive reaction is that filteringMode sounds very application layer. Our grid really is unawares of any notion of filtering

src/Hypergrid.js
joneit: always did, i believe

src/cellEditors/CellEditor.js
joneit: not sure why i removed this.... the header row count was probably out of date. i'll revisit.

src/defaults.js
Dwaynekj: Ah ok

src/cellEditors/CellEditor.js
Dwaynekj: Ok. So wait to thumbs?

joneit: The intent of the deleted line in question is to reposition the edit box before giving an error animation + alert. Not sure why that was necessary but seems to work without it. More to the point, that function `selectViewportCell ` is broken and anyway and needs to be fixed. We can do that later. This should be good to go as is.

PR #414, opened 2016-12-05
No events on blank space

Dwaynekj:
See [GRID-440](https://appoji.jira.com/browse/GRID-440)

@joneit

### DIFF
--- a/src/lib/events.js
+++ b/src/lib/events.js
@@ -241,7 +241,7 @@ module.exports = {
     delegateCanvasEvents: function() {
         var self = this;
 
-        function handleMouseEvent(e, fn) {
+        function handleMouseEvent(e, cb) {
             var primitiveEvent = self.getGridCellFromMousePoint(e.detail.mouse),
                 decoratedEvent;
 
@@ -256,9 +256,8 @@ module.exports = {
                         writable: true
                     }
                 );
-
+                cb.call(self, decoratedEvent);
             }
-           if (decoratedEvent) {fn(decoratedEvent);}
         }
 
         this.addEventListener('fin-canvas-resized', function(e) {
@@ -270,7 +269,7 @@ module.exports = {
             if (self.properties.readOnly) {
                 return;
             }
-            handleMouseEvent(e, self.delegateMouseMove.bind(self));
+            handleMouseEvent(e, self.delegateMouseMove);
         });
 
         this.addEventListener('fin-canvas-mousedown', function(e) {
@@ -288,7 +287,7 @@ module.exports = {
                 this.delegateMouseDown(mouseEvent);
                 this.fireSyntheticMouseDownEvent(mouseEvent);
                 this.repaint();
-            }.bind(self));
+            });
         });
 
         this.addEventListener('fin-canvas-click', function(e) {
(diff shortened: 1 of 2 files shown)


src/renderer/index.js
```diff
@@ -525,15 +525,19 @@ var Renderer = Base.extend('Renderer', {
...
-        // cellEvent.visibleRow = vr;
+        if (vr && vc) {
+            var mousePoint = this.grid.newPoint(x - vc.left, y - vr.top),
+                cellEvent = new this.grid.behavior.CellEvent(vc.columnIndex, vr.index);
 
-        return Object.defineProperty(cellEvent, 'mousePoint', { value: mousePoint });
+            // cellEvent.visibleColumn = vc;
+            // cellEvent.visibleRow = vr;
+
+            return Object.defineProperty(cellEvent, 'mousePoint', {value: mousePoint});
```
joneit: Can we please try whenever possible to stick to the pattern of a single `return` statement at the bottom:
```javascript
function() {
    var result;

    if (condition) {
        result = ...;
    }

    return result;
}
```
Note the falsy return value `undefined` is usually perfectly workable (unless a `null` is specifically required).

src/renderer/index.js
```diff
@@ -525,15 +525,19 @@ var Renderer = Base.extend('Renderer', {
...
+        }
+        return null;
```
joneit: Did you check _all_ usages of `getGridCellFromMousePoint ` to make sure they deal with falsy return?

src/lib/events.js
```diff
@@ -237,17 +237,24 @@ module.exports = {
...
+                    {
+                        value: e,
+                        enumerable: false,
+                        configurable: true,
+                        writable: true
+                    }
+                );
+
+            }
+           if (decoratedEvent) {fn(decoratedEvent);}
```
Dwaynekj: duh. Thanks

src/renderer/index.js
Dwaynekj: null feels more descriptive about the type signature than undefined to me

joneit: In that case the preferred pattern would be:
```javascript
function() {
    var result = null;

    if (condition) {
        result = ...;
    }

    return result;
}
```
or:
```javascript
function() {
    var result;

    if (condition) {
        result = ...;
    } else {
        result = null;
    }

    return result;
}
```

src/renderer/index.js
joneit: Nevermind, I see this call _is the only usage!_

Dwaynekj: ok

src/lib/events.js
```diff
@@ -237,17 +237,24 @@ module.exports = {
...
-                self.getGridCellFromMousePoint(e.detail.mouse),
-                'primitiveEvent',
-                {
-                    value: e,
-                    enumerable: false,
-                    configurable: true,
-                    writable: true
-                }
-            );
+        function handleMouseEvent(e, fn) {
```
Dwaynekj: Done

src/lib/events.js
Dwaynekj: Done

PR #415, opened 2016-12-06
Select cell functions

joneit:
There is no ticket associated with this PR.

Fixed vertical scroll max calculation (was not displaying final rows).

This in turn fixed `selectFinalCell` and `selectToFinalCell` which failed because final row was not displayable.

Also refined calculation of column with for handle column which was too narrow to accommodate row numbers when bolded (row selected) and was showing elipsis.

### DIFF
--- a/src/Hypergrid.js
+++ b/src/Hypergrid.js
@@ -1070,9 +1070,9 @@ var Hypergrid = Base.extend('Hypergrid', {
 
         if (
             event.isDataColumn &&
-            event.getCellProperty(event.isDataRow ? 'editable' : 'filterable')
+            event.properties[event.isDataRow ? 'editable' : 'filterable']
         ) {
-            this.setMouseDown(new Point(event.gridCell.x, event.dataCell.y));
+            this.setMouseDown(new Point(event.gridCell.x, event.gridCell.y));
             this.setDragExtent(new Point(0, 0));
 
             cellEditor = this.getCellEditorAt(event);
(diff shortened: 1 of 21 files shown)


Dwaynekj:
This brings up questions I have about [this](https://appoji.jira.com/browse/GRID-331). Since subgrids are arbitrary is this a matter of just customizing the scroll bar dom element size

joneit:
The scrollbars can be custom-positioned.

joneit:
@Dwaynekj Ready for final review on this. Let's get this sucker merged!

Dwaynekj:
Manual testing looks good

src/Hypergrid.js
```diff
@@ -2344,5 +2344,7 @@ Hypergrid.localization = {
 Hypergrid.prototype.mixIn(require('./lib/events'));
 Hypergrid.prototype.mixIn(require('./lib/selection'));
 
+Hypergrid.properties = globalProperties;
```
joneit: This is just defaults.js, which is the "shared" global properties object. Alterations here affect all grids, including simple overrides and theming. `grid.properties` is extended (created) from this and these are the only base layers now (down from 5 or 6 layers). Additions to `grid.properties` override the global props.

joneit:
@Dwaynekj Did you notice the new "Link styles" dashboard control?

src/defaults.js
```diff
@@ -255,6 +262,13 @@ var defaults = {
 
     /**
      * @default
+     * @type {string}
+     * @memberOf module:defaults
+     */
+    filterRenderer: 'SimpleCell',
```
Dwaynekj: Why was this added?

PR #421, opened 2016-12-14
GRID-452: Links

joneit:
See Jira [GRID-452](https://appoji.jira.com/browse/GRID-452) for details.

### DIFF
--- a/src/Hypergrid.js
+++ b/src/Hypergrid.js
@@ -1248,23 +1248,22 @@ var Hypergrid = Base.extend('Hypergrid', {
      * * `false` drill-down unchanged (was already in requested state)
      */
     cellClicked: function(event) {
-        var result = false;
+        var result = this.behavior.cellClicked(event);
 
-        //click occurred in background area
-        if (
-            event.gridCell.x <= this.getColumnCount() &&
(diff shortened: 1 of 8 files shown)


src/Hypergrid.js
```diff
@@ -1248,23 +1248,22 @@ var Hypergrid = Base.extend('Hypergrid', {
...
 
         return result;
     },
 
+    /**
+     * To intercept link clicks, override this method (either on the prototype to apply to all grid instances or on an instance to apply to a specific grid instance).
+     * @memberOf Hypergrid#
+     */
+    windowOpen: function(url, name, features, replace) {
+        return window.open.apply(window, arguments);
```
Dwaynekj: @joneit This should be turned off

Dwaynekj: @rdepena @datamadic @joneit We should discuss whether HG should handle links and do the actual redirect. My vote is no

Dwaynekj: Said another way. I don't think Hypergrid should call window.open or do any navigation in a window. The application user should do so

Dwaynekj:
@joneit Before this becomes tumbleweed lets bring this up with the guys tomorrow

PR #424, opened 2016-12-18
GRID 434: Hypergrid.prototype.reset

joneit:
* Generalized the `filter` and `sorter` setter/getters on grid/behavior/dataModel with `seController` and `getController` methods that take a controller `type` as a first parameter.
* The `datModel.api` hash is now `dataModel.controllers`.
* Moved the loop that finds matching data sources by (type) to `DataSourceBase`.
* Fixed `reset` by adding an `options` object and reusing previously established subgrids, pipline, and controlls when no new ones are specified.
  * Added `options` to `Hypergrid.prototype.reset` that takes:
    * `subgrids` (consumed by `Behavior.prototype.reset`)
    * `pipeline` (consumed by `dataModels.JSON.prototype.reset`)
    * `controllers` (consumed by `dataModels.JSON.prototype.reset`)
  * Added `options` to `Behavior.prototype.reset` that takes:
    * `subgrids`
    * `pipeline` (consumed by `dataModels.JSON.prototype.reset`)
    * `controllers` (consumed by `dataModels.JSON.prototype.reset`)
  * Added `options` to `dataModels.JSON.prototype.reset` that takes:
    * `pipeline` (consumed by `dataModels.JSON.prototype.setPrototype`)
    * `controllers` (consumed by `dataModels.JSON.prototype.setPrototype`)
  * Added `options` to `dataModels.JSON.prototype.setPrototype` that takes:
    * `pipeline`
    * `controllers`


Dwaynekj:
@joneit Shoudn't getController return all pipes [here ](https://github.com/openfin/fin-hypergrid-data-source-base/blob/master/index.js#L108) with the appropriate type and let the user decide?

I understand that notion of pipes itself is supposed to be a black box to the dataModel, but from the POV of people using base it won't be

add-ons/hyper-sorter/mix-ins/dataModel.js
```diff
@@ -85,7 +100,7 @@ module.exports = {
      * @desc returns the columns that currently sorted and their intended direction of the sort
      */
     getSortedColumnIndexes: function() {
-        return this.sorter.prop('sorts') || [];
+        return this.getController('sorter');
```
Dwaynekj: Assuming that defaulting to [] you are fixing in the other PR

Dwaynekj:
Finished Review. Overall I think the idea is clean. The "shape" (including controllers) of the grid remains stable unless explicitly destroyed. The other UI state pertaining to properties get wiped

demo/js/demo.js
```diff
@@ -718,9 +720,7 @@ window.onload = function() {
     });
 
     grid.addProperties({
-        readOnly: false,
-        noDataMessage: 'No Data to Display',
-        showFilterRow: true
+        readOnly: false
```
joneit: It seems to be an attempt to bail from event generation, making the grid "read-only" by ignoring all mouse and keyboard events. I'm not sure it is everywhere it needs to be. Since you're into events this week, could you make sure it is fully implemented?

joneit:
@Dwaynekj In response to [your comment](https://github.com/openfin/fin-hypergrid/pull/424#issuecomment-268046688):

`setController(type, controller)` is defined to set _all_ pipes with matching `type` to reference the (same) controller. Therefore, `getController(type)` only needs to return that one controller they all have references to.

PR #438, opened 2017-01-05
GRID-467 Treeview and other demos fixed

Dwaynekj:
…api. setPipeline must now be called before setting any controllers


src/behaviors/Behavior.js
```diff
@@ -1317,7 +1317,7 @@ var Behavior = Base.extend('Behavior', {
     },
 
     prop: function(type, columnIndex, keyOrHash, value) {
-        var result = this.dataModel.prop.apply(this.behavior, arguments);
+        var result = this.dataModel.prop.apply(this.dataModel, arguments);
```
joneit: Good catch

PR #447, opened 2017-01-11
GRID-207: Single row selection mode

joneit:
See Jira [GRID-207](https://appoji.jira.com/browse/GRID-207) for details.

This also fixes error in demo dashboard introduced by earlier PR that's throwing a console error.

### DIFF
--- a/src/Hypergrid.js
+++ b/src/Hypergrid.js
@@ -1261,22 +1261,6 @@ var Hypergrid = Base.extend('Hypergrid', {
         return window.open.apply(window, arguments);
     },
 
-    /**
-     * @memberOf Hypergrid#
-     * @desc Synthesize and dispatch a `fin-selection-changed` event.
-     */
-    selectionChanged: function() {
-        var selectedRows = this.getSelectedRows();
-        var selectionEvent = new CustomEvent('fin-selection-changed', {
-            detail: {
-                rows: selectedRows,
-                columns: this.getSelectedColumns(),
-                selections: this.selectionModel.getSelections(),
-            }
-        });
-        this.canvas.dispatchEvent(selectionEvent);
-    },
-
     getColumns: function() {
         return this.behavior.getColumns();
     },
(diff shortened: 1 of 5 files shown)


src/Hypergrid.js
```diff
@@ -1261,22 +1261,6 @@ var Hypergrid = Base.extend('Hypergrid', {
         return window.open.apply(window, arguments);
     },
 
-    /**
-     * @memberOf Hypergrid#
-     * @desc Synthesize and dispatch a `fin-selection-changed` event.
-     */
-    selectionChanged: function() {
```
Dwaynekj: Nice. Good to keep this file as lean. Even if it will still have a million API calls on it. Someday we should do the Hyperblotter revisit and API reshaping

src/lib/selection.js
```diff
@@ -364,20 +364,7 @@ module.exports = {
...
-        if (this.mouseDownState) {
-            //triggered programmatically
-            hasCTRL = this.mouseDownState.primitiveEvent.detail.primitiveEvent.ctrlKey;
-        }
-
-        result = (
-            isCheckboxOnlyRowSelections && isSingleRowSelectionMode ||
-            !isCheckboxOnlyRowSelections && (!hasCTRL || isSingleRowSelectionMode)
-        );
+        var result = this.properties.singleRowSelectionMode;
```
Dwaynekj: so what becomes of CTRL handle?

src/lib/selection.js
```diff
@@ -484,13 +471,74 @@ module.exports = {
...
+            for (var c = 0; c < colCount; c++) {
+                row[behavior.getActiveColumn(c).header] = behavior.getValue(c, topRow);
+            }
+
+            return row;
+        }
+    },
+
+    /**
+     * @memberOf Hypergrid#
```
Dwaynekj: These were straight cut & paste?

src/lib/selection.js
joneit: Just not needed here. Due to naming of `checkboxOnlyRowSelections` property, I was conflating functionality. They never seemed to "get" nor care about CTRL anyway. It does still function where needed (in other parts of code).

src/lib/selection.js
joneit: no, added lines 501..506 below

PR #454, opened 2017-01-12
Grid-456: Filter DropDown Scroll Issue

Dwaynekj:
@joneit

### DIFF
--- a/src/cellEditors/ComboBox.js
+++ b/src/cellEditors/ComboBox.js
@@ -97,6 +97,7 @@ var ComboBox = Textfield.extend('ComboBox', {
         }
     ],
 
+    // Quick Note. Fn called on init and when reposition editor for a cell still visible after scrolling
     showEditor: function() {
         var menuModesSource = this.menuModesSource,
             menuModes = this.menuModes = {};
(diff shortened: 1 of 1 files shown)


src/cellEditors/ComboBox.js
```diff
@@ -139,7 +143,7 @@ var ComboBox = Textfield.extend('ComboBox', {
 
             setModeIconAndOptgroup.call(this, toggle, mode.name, menuModes[mode.name]);
         }.bind(this));
-
+        this.grid.properties.scrollingEnabled = false;
```
joneit: This line doesn't seem to effect scrolling at all. Anyway I don't think we need to suppress scrolling at this point. The only reason we were entertaining that change was to fix the issue, but you've already fixed it.

src/cellEditors/ComboBox.js
```diff
@@ -139,7 +143,7 @@ var ComboBox = Textfield.extend('ComboBox', {
 
             setModeIconAndOptgroup.call(this, toggle, mode.name, menuModes[mode.name]);
         }.bind(this));
```
joneit: While you're in there, would you change this to the better
```javascript
   }, this);
```
thanks

src/cellEditors/ComboBox.js
Dwaynekj: Oh I forgot this

PR #462, opened 2017-01-13
GRID-358: Don't rebuild combobox dropdown

joneit:


joneit:
@dwayne moved `built = true` and now we're good

PR #464, opened 2017-01-17
Grid 201: Sorting

Dwaynekj:
Simple Sorting Demo fixed using current Pipeline layer abstraction


Dwaynekj:
@joneit Is now ready. Its just a change to a demo file, nothing in core. Please note that I had some comments in the demo file that I was going to elaborate on in a wiki. However, given our discussion to blackbox the dataModel and avoid encouraging users to use our pipeline abstraction, I think the wiki is better served as an example of how to extend `dataModel.JSON` with overrides for `getRow` and `getFields`

PR #474, opened 2017-01-20
GRID-491: Aggregations And Group View Fix

Dwaynekj:
@joneit 

We can hold off on merging this until alpha gets the ok (to avoid having to cherry pick late) *Please test locally*. You will also need this [PR](https://github.com/openfin/hyper-analytics/pull/47)

I need your help addressing these minor bugs (I'm a little tired)

* Unrelated bug in demo where clicking tree-view checkbox then clicking  "none" does not undo treeview
* Aggregations and Groupview Column have combobox editor on 

I can address this one
* Aggregations/GroupView leaves dirty state in the schema. As in the pink demo, if you do *onely* Agg/Grp on or off things work fine, but try toggling between Agg and Grp

I also got TravisCI running again

### DIFF
--- a/add-ons/aggregations-view.js
+++ b/add-ons/aggregations-view.js
@@ -104,8 +104,12 @@ AggregationsView.prototype.setAggregateGroups = function(aggregations, groups) {
     if (aggregated) {
         dataSource.setAggregateGroups(aggregations, groups);
         behavior.reindex(); // rows have changed
+        behavior.setTopTotals(behavior.dataModel.dataSource.getGrandTotals());
+        behavior.setBottomTotals(behavior.dataModel.dataSource.getGrandTotals());
     } else {
         dataSource.setAggregateGroups({}, []);
+        behavior.setTopTotals([]);
+        behavior.setBottomTotals([]);
     }
 
     behavior.createColumns(); // columns changed
--- a/src/dataModels/SummarySubgrid.js
+++ b/src/dataModels/SummarySubgrid.js
@@ -48,7 +48,7 @@ SummarySubgrid.prototype = {
     },
 
     getRow: function(y) {
-        return this.data[y];
+        return this.getData()[y];
     }
 };
 
--- a/src/dataModels/SummarySubgrid.js
+++ b/src/dataModels/SummarySubgrid.js
@@ -28,11 +28,11 @@ SummarySubgrid.prototype = {
     hasOwnData: true, // do not call setData implicitly
 
     getRowCount: function() {
-        return this.behavior.dataModel.getRowCount() && this.getData().length;
+        return this.getData().length;
     },
 
     getData: function() {
-        return this.behavior.dataModel.dataSource.getGrandTotals() || this.data;
+        return this.data;
     },
 
     setData: function(data, schema) {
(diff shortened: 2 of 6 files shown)


add-ons/aggregations-view.js
```diff
@@ -104,8 +104,12 @@ AggregationsView.prototype.setAggregateGroups = function(aggregations, groups) {
     if (aggregated) {
         dataSource.setAggregateGroups(aggregations, groups);
         behavior.reindex(); // rows have changed
+        behavior.setTopTotals(behavior.dataModel.dataSource.getGrandTotals());
```
joneit: These lines should not be necessary. See lines [16](https://github.com/openfin/fin-hypergrid/blob/develop/src/dataModels/SummarySubgrid.js#L16) and [35](https://github.com/openfin/fin-hypergrid/blob/develop/src/dataModels/SummarySubgrid.js#L35) of SummarySubgrid.js.

src/dataModels/SummarySubgrid.js
```diff
@@ -28,11 +28,11 @@ SummarySubgrid.prototype = {
     hasOwnData: true, // do not call setData implicitly
 
     getRowCount: function() {
-        return this.behavior.dataModel.getRowCount() && this.getData().length;
+        return this.getData().length;
     },
 
     getData: function() {
-        return this.behavior.dataModel.dataSource.getGrandTotals() || this.data;
+        return this.data;
```
joneit: Why did you change this? What was the problem? Let's change this back, please.

src/dataModels/SummarySubgrid.js
```diff
@@ -28,11 +28,11 @@ SummarySubgrid.prototype = {
     hasOwnData: true, // do not call setData implicitly
 
     getRowCount: function() {
-        return this.behavior.dataModel.getRowCount() && this.getData().length;
+        return this.getData().length;
```
joneit: Good catch.

src/dataModels/SummarySubgrid.js
Dwaynekj: Despite the fact that this plugin won't be in core anymore, hence not of major concern, why are we reaching into one of a pipes for a method? The subgrids should not call non datasource-base interface methods

This is why I added these [line](https://github.com/openfin/fin-hypergrid/pull/474/files#diff-2a01e55e2f6e71aa32f816c37d8e76b7R107)

Dwaynekj: As per discussion I will changed to this 
`return this.data.length ? this.data : this.behavior.dataModel.dataSource.getGrandTotals();`

joneit: Don't worry about this for now; I'll do it with my subgrids PR... I also fixed `setValue` which is broke (not in use; never tested).

Dwaynekj: Ok I'll disregard this change

src/dataModels/SummarySubgrid.js
Dwaynekj: Done

add-ons/aggregations-view.js
Dwaynekj: I'll let you make the change on your branch

PR #482, opened 2017-01-31
GRID-455-459-475-470: Resize last col & Drag Cell Selectionm & Dragging Last Column

Dwaynekj:
@joneit I intend to push an alpha tonight that won't be a 1.3.0 master target

PR #486, opened 2017-02-01
GRID-465: Hide hover effects when mouse leaves canvas element

joneit:
Another quickie!


Dwaynekj:
Can you look at https://github.com/openfin/fin-hypergrid/pull/482 post merging this. This may be the final piece missing in that item

PR #491, opened 2017-02-03
GRID-495: Row selections after filter cell edit

joneit:
Fixed per Jira tkt.


Dwaynekj:
@joneit Could describe the changes. Whats the new behavior

PR #497, opened 2017-02-07
Pseudo-require for Hypergrid (build file only)

joneit:


src/index.js
```diff
@@ -29,6 +29,56 @@ Hypergrid.rectangular = require('rectangular');
 Hypergrid.lib = require('./lib');
 Hypergrid.Base = require('./Base');
 
+var windowRequire = window.require;
+
+window.require = function(path) {
```
Dwaynekj: Concerned about doing this.

src/index.js
```diff
@@ -29,6 +29,56 @@ Hypergrid.rectangular = require('rectangular');
...
+    var result, crumbs, i,
+        errMsg = 'Path ' + path + ' unknown or not exposed in build file.';
+
+    if (path.indexOf('fin-hypergrid/') === 0) {
```
Dwaynekj: Yes I see that you using your require when the path is local to Hypergrid otherwise theirs. I still have concerns about this. Lets discuss after merge to master

src/index.js
joneit: This does not affect commonJS/Browserify usage as that `require` is a passed parameter.

If developer already had a definition for `window.require` by the time `fin-hypergrid.js` is loaded (via `<script>` tag), that definition is saved in a local var (`windowRequire`), and any call to `require(file)` where `file` does not start with 'fin-hypergrid/' calls the saved function instead. Otherwise, it returns the object exposed via property of `fin.Hypergrid` constructor.

So previously, there were two different approaches to referencing exposed objects:

```javascript
var Behavior = fin.Hypergrid.behaviors.Behavior; // if using build file
var Behavior = require('fin-hypergrid/src/behaviors/Behavior'); // if using npm module
```

This is a source of confusion, needs effort to repurpose existing code to Browserify, and forces me to have to show both ways in every example I write for wikis etc.

Now with this new approach we simply use `require` in all cases.

joneit:
This latest commit "privatizes" the function.

PR #499, opened 2017-02-07
Grid-459: Unable to resize Column

Dwaynekj:
Also fixes an error in console that occurs when 

Scenario A
1) Load Hypergrid Demo.
2) Click on Dashboard and click "Set Data 1(5000 rows)". To get more columns on the grid.
3) Scroll Horizontally so move to few columns to the right.
4) Now click on "FilterCell" and press Enter. 
5) You can see exception in console saying "columnIndex" not found.

Scenario B
1) Load Hypergrid Demo.
2) Apply a filter so that there are no records on the grid.
3) Now go back to the filter cell and press enter. 
4) You can see exception in console saying "rowIndex" not found.

### DIFF
--- a/demo/js/demo.js
+++ b/demo/js/demo.js
@@ -166,7 +166,7 @@ window.onload = function() {
 
     var gridOptions = {
             data: people1,
-            margin: { bottom: '17px' },
+            margin: { bottom: '17px', right: '17px'},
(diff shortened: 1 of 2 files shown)


Dwaynekj:
@joneit Feel free to review post-merge. William had caught error where we incorrectly getting the column index value from the visible columns array

Issue #533, opened 2017-02-15
Empty rows when scrolling in Groupview


Dwaynekj:
GROUPVIEW
When we expand few groups in groupview and scroll vertically. We observe that the rows are getting empty when scrolling at a particular location of the grid.
Load the Demo
Switch to GroupView
Expand group 'Alaska' and 'Arkansas'.
Now try scrolling vertically. 
You can see the grid rows are empty as they cross certain location of the grid.

PR #538, opened 2017-02-23
ISSUE#533: Empty Rows when Scrolling

Dwaynekj:
@joneit This only affects group-view presentation


Dwaynekj:
Resolves https://github.com/openfin/fin-hypergrid/issues/533

Issue #530, opened 2017-02-11
Multiple-Grids demo


willsolve:
I'm getting the following error with the multiple-grids.html demo:
`Canvas.js:284 TypeError: Cannot read property 'length' of undefined
    at Constructor.paintCellsByColumnsAndRows [as paintCells] (file:///C:/Users/user/Desktop/demo/fin-hypergrid/build/fin-hypergrid.js:24702:19)
    at Constructor.renderGrid (file:///C:/Users/user/Desktop/demo/fin-hypergrid/build/fin-hypergrid.js:25752:14)
    at Constructor.paint (file:///C:/Users/user/Desktop/demo/fin-hypergrid/build/fin-hypergrid.js:25473:18)
    at Canvas.paintNow (file:///C:/Users/user/Desktop/demo/fin-hypergrid/build/fin-hypergrid.js:20009:28)
    at Canvas.tickPaint (file:///C:/Users/user/Desktop/demo/fin-hypergrid/build/fin-hypergrid.js:19895:18)
    at Canvas.beginPainting.tickPainter (file:///C:/Users/user/Desktop/demo/fin-hypergrid/build/fin-hypergrid.js:19913:18)
    at file:///C:/Users/user/Desktop/demo/fin-hypergrid/build/fin-hypergrid.js:20371:27
    at Array.forEach (native)
    at paintLoopFunction (file:///C:/Users/user/Desktop/demo/fin-hypergrid/build/fin-hypergrid.js:20369:20)`

The problem happens when this is called on the second iteration/instance
`grid[i] = new fin.Hypergrid(options);`

I traced it to the following code, and if the reset line is commented out, then it renders again:
fin-hypergrid.js line 24691 (by-columns-and-rows.js line 52)

   if (paintCellsByColumnsAndRows.reset) {
        this.resetAllGridRenderers();
/*        paintCellsByColumnsAndRows.reset = false; */
        bundleRows.call(this, false);
        bundleColumns.call(this, true);
    } else if (paintCellsByColumnsAndRows.rebundle) {
        paintCellsByColumnsAndRows.rebundle = false;
        bundleColumns.call(this);
    }

Dwaynekj:
@willsolve Thanks for pointing this out. We will add to our backlog

Dwaynekj:
[...]

PR #564, opened 2017-03-01
GRID 504: Mulit-grid demo reset bug

joneit:
The "proper" solution would be to make the grid renderers objects to contain their `reset` and `rebundle` flag states. Instead, this simple solution moves those from the function objects to the `gridRenderers` hash in renderer/index.js.

@Dwaynekj, what to do you think? Either way requires changes to all grid renderers (paintCells functions). The solution that would recast them as objects would perhaps be clearer, but it's a heavier solution (more edits, more memory, more code).


Dwaynekj:
Reference issue https://github.com/openfin/fin-hypergrid/issues/530

Dwaynekj:
@joneit LMK When you have tested it. We can plan on reviewing it Monday

PR #580, opened 2017-04-03
GRID-373: Grid state part 3 - calculators

joneit:
Column calculators now share functions, and can be either a function, a name of a function, a named function in a string, or an anonymous function in a string. See Wiki-to-come for examples.

Task list:
- [x] `rows` _(merged in part 1)_
- [x] `columns` _(merged in part 1)_
- [x] `cells` _(merged in part 1)_
- [x] `columnNames`
- [x] `subgrids`
- [x] `calculators`
- [ ] `plugins`
- [x] `Hypergrid.saveState` and `loadState`
- [ ] accept a state object on grid instantiation (or allow the current `options` object to be a state object)


Dwaynekj:
@joneit Is this ready for review

joneit:
Yes, was ready for review. I just made 1 cosmetic change. It may be hard to appreciate what I've done here without the examples in the wiki I'm working on.

joneit:
Added `Hypergrid.loadState` and `.saveState`.

joneit:
Added `Hypergrid.loadState` and `.saveState` but the latter only saves `columns` and `rows` so far... needs more work... **but mergeable as is.**

Dwaynekj:
@joneit Did you see my comment about the functions? I thought we were just going to store the function name

joneit:
@Dwaynekj, **I do _not_ see a comment from you here, nor in slack, nor in the Jira ticket, nor in email.** Details of the expanded design follow.

#### Abstract

I arrived at a sweeping but simple design, backwards-compatible to the current practice, which is that the column's `calculator` property is a function reference, which is the most performant. What bdbd4b7  adds is that we now also accept function _names;_ and going forward that will be the recommended practice.

#### Brief review of previous functionality

`calculator` is defined as a dynamic _column-only_ property, which means it is a getter/setter that resides in `behaviors/columnProperties.js` rather than in `lib/dynamicProperties.js` (which is for grid-level properties).

The previous definition of the `calculator` setter was simply that it ran the [R-value[(https://en.wikipedia.org/wiki/Value_(computer_science)#Assignment:_l-values_and_r-values) (new value) through `toFunction` to create functions out of strings. (`toFunction` is a no-op when the value is already a function.)

The function reference is then copied in turn, by the `calculator` setter defined in Column.js, to the schema, which serves as the actual backing store for the getter, and which provides access to the calculator from the data models.

#### New functionality

This ticket expands the `calculator` setter in `behaviors/columnProperties.js` to do two more things:
1. It now accepts a function name as well, which it uses to find the actual function in the registry, passing along a function reference, as before.
2. Any functions given here are now immediately copied to the registry using the name in the function as the registry key; or if the function has no name, the serialized function itself is used as the registry key.

The first thing to notice about the above is that columns that share a function will now share a reference to a single instance of that function, whereas they were previously independent. That is the is the other big advantage of having the registry (besides getting the functions out of the state).

#### The registry

The application developer has the choice of providing the calculator functions along with his application code, reading them in separately via JSON or JSON-p, or allowing them to be read in as part of the state JSON (in a root-level hash, `grid.properties.calculators`) if it doesn't bother him to mix data and code together. (Some people are more religious about this than others.)

The code expects the registry to be in the grid properties. If the developer wants to persist it separately, he needs to set `grid.properties.calculators` to point to it before setting state. The `saveState` method has the option of blacklisting this property to prevent it from being included in the persisted state serialization.

#### Summary

The value given with the `calculator` key may be:
* a function name, which is resolved using the registry
* a named function, which is added to the registry using the function name as the registry key
* a named function in a string, which is run through `toFunction`, and added to the registry using the function name as the registry key
* an anonymous function, which is added to the registry as is, using its serialized form as the registry key
* an anonymous function in a string, which is run through `toFunction`, and added to the registry using its serialized form as the registry key

In _all cases,_ the result is function a reference; and if functions were given (in any form), they are added to the registry as needed (_i.e.,_ when missing).

src/Hypergrid.js
```diff
@@ -2190,6 +2205,15 @@ function setStyles(el, style, keys) {
     }
 }
 
+function stringifyFunctions() {
+    var self = this;
+    return Object.keys(this).reduce(function(obj, key) {
```
joneit: As explained above, this design encompasses a map. It uses a map if provided; and constructs a map if not provided. The next saveState will then persist a map.

joneit:
@Dwaynekj I don't see your alpha release. I propose merging this PR and creating a new alpha...

Dwaynekj:
This is the alpha release https://github.com/openfin/fin-hypergrid/tree/v1.4.0-alpha

Feel free to merge and create a new release

PR #584, opened 2017-04-12
GRID-373: Grid state, part 4

joneit:
This PR:
1. Moves the dynamic grid props from `grid.theme` to `grid.properties`. This internal change simplifies `saveState`. _(Details below.)_
2. Implements the last item in the checklist, accepting a state object on grid instantiation.

#### Task list:
- [x] `rows` _(merged in part 1)_
- [x] `columns` _(merged in part 1)_
- [x] `cells` _(merged in part 1)_
- [x] `columnNames`
- [x] `subgrids`
- [x] `calculators`
- [ ] `plugins` _(Not sure we're going to do this.)_
- [x] `Hypergrid.saveState` and `loadState`
- [x] accept a state object on grid instantiation (or allow the current `options` object to be a state object)

#### Regarding moving dynamic grid props
_The original design placed the "dyna props" (getters/setters) between the defaults layer and the theme layer. This dynamic props layer would be in the trunk (one instance for all grid instances). However, I eventually decided that for performance sake it wasn't worth adding an extra layer, and added the dyna props instead to the theme layer. At the time there was one global theme, which assumed that all grid instances would want to have the same look. I eventually created a separate theme layer for each grid. This had the side effect of having to create the dyna props anew for each grid instance. No biggie since the functions in the property descriptors object are shared. However, at that point there was no longer any point in having them in any other layer than the grid props layer itself. This was not important until `saveState` had to create proxies in the grid props layer to get `JSON.stringify` to output them. Moving them to their rightful position obviated this need. It also makes clearing the theme layer simpler._


joneit:
@Dwaynekj In your absence I'm going to merge this without further review; and create a new alpha (1.4.2).

PR #597, opened 2017-05-22
Update scroll limits when adding new properties to the grid (defaultRowHeight may have changed)

hugoeanogueira:
Without this, when instantiating a new grid and then adding theming properties with addProperties(),
the scroll limits are not updated and only a portion of the grid may be accessible on scroll.


Dwaynekj:
@hugoeanogueira thanks!

Issue #519, opened 2017-02-08
Row Selection and Column Selection Keyboard Nav not working


Dwaynekj:

hugoeanogueira:
Hi! Not sure if this bug is related with the problem i've found but here it goes. 
While testing the new 2.0.0 release, i've noticed that the keyboard navigation was broken, noticeable by the following error in chrome dev tools:

```
Uncaught TypeError: Cannot read property 'getCellOwnProperties' of undefined
    at CellEvent.get (cellEventFactory.js:56)
    at CellEvent.get (cellEventFactory.js:61)
    at Constructor.handleKeyDown (CellSelection.js:112)
    at Constructor.handleKeyDown (Filters.js:34)
    at Constructor.onKeyDown (Behavior.js:916)
    at Constructor.delegateKeyDown (events.js:579)
    at events.js:456
    at HTMLCanvasElement.decorator (events.js:30)
    at Canvas.dispatchNewEvent (Canvas.js:311)
    at Canvas.finkeydown (Canvas.js:451)
```
This does not affect v1.3.0, but was already present in 2.0.0-alpha.

From the demo below, we can see that this somehow related with the grid being cropped in the viewport (having more columns that it can show).
For some reason, when it tries to handle the 'keydown' event by retrieving `cellOwnProperties`, the `CellEvent` looks empty, without the `column` property and `this.column.getCellOwnProperties` throws up.

```
<!DOCTYPE html>
<html>
<head>
  <meta charset="utf-8">
  <meta name="viewport" content="width=device-width">
  <title>JS Bin</title>
  <script type="text/javascript" src="https://openfin.github.io/fin-hypergrid/build/fin-hypergrid.js"></script>
</head>
<style type="text/css" media="screen">
  .grid {
    width: 800px;
    height: 220px;
    margin-bottom: 12px;
    background-color: #fafafa;
  }
</style>
<body>
  <h1>Keyboard navigation bug demo</h1>
  <div id="smallGrid" class="grid"></div>
  <div id="bigGrid" class="grid"></div>

  <script type="text/javascript">
    function createRow(text, cols) {
      var arr = {};
      for(var i = 0; i < cols; i++) {
        arr['key_' + i] = text;
      }
      return arr;
    }

    var smallGrid = new fin.Hypergrid(document.getElementById('smallGrid'), {
      data: [
        createRow('this', 10),
        createRow('works', 10),
        createRow('great', 10),
        createRow('with', 10),
        createRow('a', 10),
[...]

Issue #515, opened 2017-02-08
Unable to scroll to last column


Dwaynekj:
When we have more than 15 visible columns, we unable to scroll to last column, last row of the grid by using below API.
g.selectCellAndScrollToMakeVisible(g.getColumnCount() - 1, g.behavior.dataModel.getRowCount() - 1)

Steps to reproduce
1) Launch Hypergrid demo
2) Click on Dashboard and click 'Set Data 1 (5000 rows)'
3) Reduce the widths of few columns such way that you have more than 17 columns on the viewport.
4) Now click on the first cell on the view port (0,0) by clicking on it.
5) Open console and execute below command to move to the last cell of the grid.
g.selectCellAndScrollToMakeVisible(g.getColumnCount() - 1, g.behavior.dataModel.getRowCount() - 1)
6) You can see that it does not bring the last cell into the viewport.

Dwaynekj:
@joneit I think this lies in how computeCellBounds builds the list of visibleColumns starting from the left where as the scrollH value diff is based on the right renderer.dataWindow.corner. The visibleColumns array maxes out based on available grid width before it even considering the column it was trying to show. I don't think this is an elegant solution, but it seems to prove this theory

https://github.com/openfin/fin-hypergrid/pull/541/files

PR #598, opened 2017-05-24
fix(#519): fix keyboard navigation

hugoeanogueira:
This PR should fix multiple issues regarding the keyboard navigation.

1) Basic keyboard navigation wasn't working. 
2) Expanding selection with keyboard was broken when selection got bigger than viewport (either vertically or horizontally).
- Bug example with horizontal scrolling:
fixed_bug4_1
- Bug example with vertical scrolling:
fixed_bug4_2

3) Expanding selection vertically with keyboard was bugged when reaching the top of the grid
fixed_bug1

### DIFF
--- a/src/Hypergrid.js
+++ b/src/Hypergrid.js
@@ -1255,12 +1255,11 @@ var Hypergrid = Base.extend('Hypergrid', {
             // target is to left of scrollable columns; negative delta scrolls left
             if ((delta = c - dw.origin.x) < 0) {
                 this.sbHScroller.index += delta;
-            }
 
             // target is to right of scrollable columns; positive delta scrolls right
             // Note: The +1 forces right-most column to scroll left (just in case it was only partially in view)
-            if ((c - dw.corner.x + 1) > 0) {
-                this.sbHScroller.index = this.getMinimumLeftPositionForColumn(c);
+            } else if ((c - dw.corner.x + 1) > 0) {
+                this.sbHScroller.index = this.getMinimumLeftPositionToShowColumn(c);
             }
         }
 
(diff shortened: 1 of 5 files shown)


src/Hypergrid.js
```diff
@@ -1257,7 +1258,7 @@ var Hypergrid = Base.extend('Hypergrid', {
 
                 // target is to right of scrollable columns; positive delta scrolls right
                 // Note: The +1 forces right-most column to scroll left (just in case it was only partially in view)
-                (delta = c - dw.origin.x + 1) > 0
```
Dwaynekj: my initial concern is how this affects,  https://github.com/openfin/fin-hypergrid/issues/515. Its a uses case we have for one of our customers

hugoeanogueira: Yup, looks like its affected...

hugoeanogueira:
The scroll to last column is bugged with these changes....

Apart from that, I'm having some troubles to replicate the other bugs with the released 2.0.0 version (I can replicate them in our project, but not in a new standalone project). I'll try to fix them pointing already to this PR.
To be aware, here's the list:
- When on the bottom of the grid, while expanding selection with keyboard, the SHIFT+UP expands selection correctly, but with SHIFT+DOWN the selection region disappears.
bug2_invisible_selection

- Selection regions pass over column headers while scrolling.
bug3_selection_overlay

- When we type in a cell, which is outside viewport, it crashes (should scroll into view and open editor?)
bug5_editor_scroll

- If we navigate until the bottom of the grid, then going up until some scroll is done, and finally navigate until the bottom, the last row is cropped (depends on viewport size).
bug6_scroll_crop

src/Hypergrid.js
Dwaynekj: Any thoughts on how we support both?

Dwaynekj: Maybe split the code path for the features, and be less DRY?

hugoeanogueira: Don't know... I'll try to look at this today.

hugoeanogueira: Done! :p
Pull the last changes (2 commit I think...) and test it again please.

I've updated the scrollToMakeVisible() code so that, only when it needs to scroll right, it calculates each columns left and right positions, until reaching the target column's right position. Knowing this, we then calculate the available viewport dimensions (without any fixed columns and headers width) and check which is the minimal index that we can scroll into, that will still show the target column fully inside the viewport. 
I've done some tests with multiple column's configurations (e.g. widths, fixed position, without numbers column) and looks good.

Hope it works!
Thanks!

hugoeanogueira:
Hey @Dwaynekj!
Not sure if you've noticed, but i've made some changes to fix this.
Looks good to you?
Thanks!

Dwaynekj:
Awesome! I'll review tomorrow

Dwaynekj:
@hugoeanogueira Just to be sure, your saying you fix the issue you reported as well as the one I brought up?

hugoeanogueira:
Yup, both the basic keyboard navigation and the selectLastCell method.

Dwaynekj:
@hugoeanogueira Sorry for the delay. I had alot of meetings today. I'll have this reviewed by Friday. LMK if this is an issue. Really appreciate your help

hugoeanogueira:
Hi. Any news @Dwaynekj?
Thanks!

Dwaynekj:
@hugoeanogueira Sorry for the delay. I'll have an update tomorrow. I've been cleaning up the old add-ons for a customer who is requested them

Dwaynekj:
@hugoeanogueira I'm starting your PR review

Dwaynekj:
@hugoeanogueira Staring your PR review!

src/Hypergrid.js
```diff
@@ -1277,6 +1277,66 @@ var Hypergrid = Base.extend('Hypergrid', {
         }
     },
 
+    /**
+     * @desc Calculate the minimum left column index so the target column shows up in viewport (we need to be aware of viewport's width, number of fixed columns and each column's width)
+     * @param {number} targetColIdx - Target column index
+     * @returns {number} Minimum left column index so target column shows up
+     */
+    getMinimumLeftPositionToShowColumn: function(targetColIdx) {
```
Dwaynekj: @hugoeanogueira Could you move this function to grid.renderer

src/Hypergrid.js
```diff
@@ -1151,7 +1151,7 @@ var Hypergrid = Base.extend('Hypergrid', {
      * @returns {boolean} The given row is fully visible.
      */
     isDataRowVisible: function(r) {
-        return this.renderer.isRowVisible(r);
+        return this.renderer.getVisibleDataRow(r);
```
Dwaynekj: @hugoeanogueira Are you suggesting here that isRowVisible needs to be subgrid aware?

hugoeanogueira: To be honest, I don't really remember why I've changed this (probably was a first attempt to fix it...) and why I didn't use the `this.renderer.isVisibleDataRow(r)` instead.
I've reverted this code and tested. Keyboard navigation, selection and the method to focus in one particular cell look good.

src/Hypergrid.js
hugoeanogueira: Yup, done!

Dwaynekj:
@hugoeanogueira I will have more updates tomorrrow

Issue #621, opened 2017-08-27
First character is wrongly mapped for non-US keyboards when opening the cell editor (editOnKeyDown=true)


hugoeanogueira:
**Description**
When using the keyboard to open cell editor, by start typing a visible character, some characters may not be correctly mapped in the `charMap` from `Canvas.js` when the user is using a keyboard layout different from the US (e.g. portuguese layout).

**Test conditions**
OS: Mac OS v10.12.6
Browsers: Chrome 60.0.3112.113 and Firefox 55
Keyboard layout: Portuguese

**How to reproduce**
- Set the portuguese keyboard layout in OS settings (e.g. mac);
- Open hypergrid's demo app;
- Select an editable cell with a simple click (cell must stay active, with editor closed);
- Type `_` (I think it corresponds to SHIFT + 0) multiple times;
- Check if all the typed characters are the `=`, e.g. `====`, or if a wrong character was typed in the beginning like this `)===`.

**Demo**
- [PT keyboard](https://user-images.githubusercontent.com/4893840/29754041-abe2477c-8b75-11e7-858f-4ea05682f661.gif)
- [US keyboard](https://user-images.githubusercontent.com/4893840/29754043-b07b5b66-8b75-11e7-83cb-27e9b754c0cc.gif)

joneit:
Resolved by **Hypergrid v2.2.11** (PR #722).

Issue #522, opened 2017-02-08
Reshape fin-* events 


Dwaynekj:
This [branch](https://github.com/Dwaynekj/fin-hypergrid/tree/IMPRV/GRID-400)

### Remaining work

- [ ] Provide backwards compatibility by preserving addEventListener and recommending addEventListenerV2
- [ ]  Provide two copies of canvas.js one of which emits the old deprecated fin-canvas events
- [ ] Provide two copies of events.js one of which dispatches the old public shape but reshapes the internal event handlers on the feature chain to use the new shape

### Summary of whats done thus far.

* Got rid of all fin-canvas* events  …
* Internal and external event handlers now utilize the same event shape
* Fixed issues with mouse events dragged off canvas
* The new event shape available is 

```
e.detail
    primitive - Original Canvas Event
    detail may be a Cell Event
    if the event was mouse or keyboard related
    mouse and keys are available.
```

#### Alternatively

```
e.detail
  primitiveEvent
  cellEvent
  mouse 
  keys
```

PR #622, opened 2017-08-27
fix(#621) - fix first typed character when opening the cell editor for non-US keyboards

hugoeanogueira:
**Description**
When using the keyboard to open cell editor, by start typing a visible character, some characters may not be correctly mapped in the charMap from Canvas.js when the user is using a keyboard layout different from the US (e.g. portuguese layout).

I've updated the `getKeyChar()` to prevent using the mapped key if it's a simple character (e.g. a letter, a digit or a symbol). Because there's too much logic in hypergrid that depends on the complex keys returned here, e.g. keys for arrows and some keys combined with shift, we cannot remove the charmap and just return the original `e.key`.
The charMap can be stripped down of simple keys and `getKeyChar()` refactored to fallback to `e.key` every time a mapping returns `undefined`.

If #522 goes forward, the code may be refactored to rely more on the original keyboard and mouse event data , like `e.key` values, instead of the mapped keys.

**Test conditions**
OS: Mac OS v10.12.6
Keyboard layout: Portuguese and US international

### DIFF
--- a/src/lib/events.js
+++ b/src/lib/events.js
@@ -112,7 +112,7 @@ module.exports = {
         return dispatchEvent.call(this, 'fin-editor-keyup', {
             input: inputControl,
             keyEvent: keyEvent,
-            char: this.canvas.getCharMap()[keyEvent.keyCode][keyEvent.shiftKey ? 1 : 0]
+            char: this.canvas.getKeyChar(keyEvent)
         });
     },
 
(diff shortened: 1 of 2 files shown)


src/lib/events.js
```diff
@@ -112,23 +112,23 @@ module.exports = {
         return dispatchEvent.call(this, 'fin-editor-keyup', {
             input: inputControl,
             keyEvent: keyEvent,
-            char: this.canvas.getCharMap()[keyEvent.keyCode][keyEvent.shiftKey ? 1 : 0]
+            char: this.canvas.getKeyChar(keyEvent)
```
joneit: 👍

Dwaynekj:
@hugoeanogueira @joneit Whats the status here

joneit:
Closing. Worked into **Hypergrid v2.2.11** (PR #722).

Issue #623, opened 2017-08-28
Space key does not open cell editor


hugoeanogueira:
**Description**
When using the keyboard to start cell editor, we cannot use the space key to open the cell editor and replace its content with a new space character.
Everything works fine if we type a "normal" character.

**Test conditions**
OS: Mac OS v10.12.6
Browsers: Chrome 60.0.3112.113 and Firefox 55

**How to reproduce**
- Open hypergrid's demo app;
- Focus on an editable cell (either with or without content);
- Hit the "space" key (nothing happens -> should have opened the editor);
- Hit the "a" key (a new editor with "a" value type should be rendered).

joneit:
Resolved by **Hypergrid v2.2.11** (PR #722).

PR #624, opened 2017-08-28
fix(#623) - Fix open editor with space key (must be treated as a normal character)

hugoeanogueira:
Fixes [#623](https://github.com/fin-hypergrid/core/issues/623).

**Description**
When using the keyboard to start cell editor, we cannot use the space key to open the cell editor and replace its content with a new space character.
Everything works fine if we type a "normal" character.

**Test conditions**
OS: Mac OS v10.12.6

**How to reproduce**
- Open hypergrid's demo app;
- Focus on an editable cell (either with or without content);
- Hit the "space" key (nothing happens -> should have opened the editor);
- Hit the "a" key (a new editor with "a" value type should be rendered).

### DIFF
--- a/src/features/CellEditing.js
+++ b/src/features/CellEditing.js
@@ -3,6 +3,15 @@
 var Feature = require('./Feature');
 var CellEditor = require('../cellEditors/CellEditor');
 
+var KEYS = {
+    RETURN: 'RETURN',
+    RETURNSHIFT: 'RETURNSHIFT',
+    DELETE: 'DELETE',
+    BACKSPACE: 'BACKSPACE',
+    SPACE: 'SPACE',
+    F2: 'F2'
+};
+
 /**
  * @constructor
  * @extends Feature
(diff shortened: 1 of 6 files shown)


src/features/CellEditing.js
```diff
@@ -3,12 +3,39 @@
...
+};
+
 /**
  * @constructor
  * @extends Feature
  */
 var CellEditing = Feature.extend('CellEditing', {
 
+    isVisibleChar: function(char, event) {
+        return !!(
```
joneit: `!!` pointless here

src/features/CellEditing.js
```diff
@@ -42,23 +69,22 @@ var CellEditing = Feature.extend('CellEditing', {
      * @memberOf KeyPaging.prototype
      */
     handleKeyDown: function(grid, event) {
-        var char, isVisibleChar, isDeleteChar, editor, cellEvent;
+        var char = event.detail.char,
+            isVisibleChar = this.isVisibleChar(char, event),
+            isSpaceChar = this.isSpaceChar(char, event),
+            isDeleteChar = this.isDeleteChar(char, event),
+            isEditChar = this.isEditChar(char, event),
+            isValidChar = !!(isVisibleChar || isSpaceChar || isDeleteChar || isEditChar),
```
joneit: `!!` pointless here

src/features/CellEditing.js
```diff
@@ -3,12 +3,39 @@
...
+    F2: 'F2'
+};
+
 /**
  * @constructor
  * @extends Feature
  */
 var CellEditing = Feature.extend('CellEditing', {
 
+    isVisibleChar: function(char, event) {
```
joneit: 1. Is it your intention to add to the `CellEditing` public interface? Because if not, we prefer local functions called with `.call(this,`.
2. making these four into functions is kind of pointless when called once, but OK, for sake of making these logic methods overridable

src/features/CellEditing.js
```diff
@@ -3,12 +3,39 @@
...
+    isSpaceChar: function(char, event) {
+        return char === KEYS.SPACE;
+    },
+
+    isDeleteChar: function(char, event) {
+        return (char === KEYS.DELETE || char === KEYS.BACKSPACE);
+    },
+
+    isEditChar: function(char, event) {
+        return (char === KEYS.F2 || char === KEYS.RETURN || char === KEYS.RETURNSHIFT);
```
joneit: including RETURN here opens the editor but short circuits current functionality which is that RETURN goes down one cell and SHIFT-RETURN goes up one cell (as down- and up-arrrow, respectively). Let's remove these from here. Developer could override this method to get this functionality.

src/features/CellEditing.js
```diff
@@ -42,23 +69,22 @@ var CellEditing = Feature.extend('CellEditing', {
...
-            )
-        ) {
+        if (isEditable && isValidChar) {
             editor = grid.onEditorActivate(cellEvent);
 
             if (editor instanceof CellEditor) {
-                if (isVisibleChar) {
-                    editor.input.value = char;
+                if (isVisibleChar || isSpaceChar) {
+                    editor.input.value = isSpaceChar ? ' ' : char;
```
joneit: better:
```js
if (isSpaceChar) {
    editor.input.value = ' ';
} else if (isVisibleChar) {
    editor.input.value = char;
```

src/features/CellEditing.js
hugoeanogueira: No, there's no need to expose this (i've only created the functions to simplify handleKeyDown). I'll update this.

src/features/CellEditing.js
hugoeanogueira: Yup, it'll be removed when the code is refactored back to handleKeyDown.

src/features/CellEditing.js
hugoeanogueira: +1

src/features/CellEditing.js
hugoeanogueira: +1

hugoeanogueira:
Closed because i've messed up with versioning... --'
Opening a fresh PR [#626 here](https://github.com/fin-hypergrid/core/pull/626/files).

PR #626, opened 2017-09-04
fix(#623) - Fix open editor with space key (must be treated as a normal character)

hugoeanogueira:
Fixes [#623](https://github.com/fin-hypergrid/core/issues/623).

**Description**
When using the keyboard to start cell editor, we cannot use the space key to open the cell editor and replace its content with a new space character.
Everything works fine if we type a "normal" character.

**Test conditions**
OS: Mac OS v10.12.6

**How to reproduce**
- Open hypergrid's demo app;
- Focus on an editable cell (either with or without content);
- Hit the "space" key (nothing happens -> should have opened the editor);
- Hit the "a" key (a new editor with "a" value type should be rendered).

Note: I've created this new PR because I've messed up with the original branch history.

### DIFF
--- a/src/features/CellEditing.js
+++ b/src/features/CellEditing.js
@@ -3,11 +3,38 @@
 var Feature = require('./Feature');
 var CellEditor = require('../cellEditors/CellEditor');
 
+var KEYS = {
+    RETURN: 'RETURN',
+    RETURNSHIFT: 'RETURNSHIFT',
+    DELETE: 'DELETE',
+    BACKSPACE: 'BACKSPACE',
+    SPACE: 'SPACE',
+    F2: 'F2'
+};
+
 /**
  * @constructor
  * @extends Feature
  */
 var CellEditing = Feature.extend('CellEditing', {
+
+    isVisibleChar: function(char, event) {
+        return !!(
+            (char.length === 1) && !(event.detail.meta || event.detail.ctrl)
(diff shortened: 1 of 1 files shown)


hugoeanogueira:
@joneit Hi! I've applied the changes suggested in the previous code review.
Sorry for creating this new PR, but I've messed up a bit with the previous branch history and this should fix it.

Regarding your comment about the RETURN and RETURNSHIFT code:
> including RETURN here opens a cell's editor on receipt, but short circuits current functionality which is that RETURN goes down one cell and SHIFT-RETURN goes up one cell (like down- and up-arrrow, respectively). Let's remove these from here. Developer could override this method to get this functionality.

The RETURN keys are handled before this code, if they're mapped as navigation keys.
Because the handler (`CellSelection.handleKeyDown`) prevents their propagation, I've removed the bit of code that was checking for it in line 60.
BTW we've have a local branch that adds the feature to use RETURN and RETURNSHIFT to open cell editor, but keeping them as navigation keys, so they can close the editor and execute the move down or up code. This feature can be toggled on/off with the `editOnReturnKeydown` property.

Dwaynekj:
@hugoeanogueira @joneit Whats the status here

joneit:
Closing. Worked into **Hypergrid v2.2.11** (PR #722).

PR #637, opened 2017-11-27
Migrate eslint from v1 to v4

joneit:
Hi guys @Dwaynekj @dcchuck This is a straight-forward PR to update eslint.
btw, I notice we now have squash merge set. Was this intentional? I would be happy to add my 2¢ to this decision.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -21,10 +21,7 @@
       2,
       "never"
     ],
-    "space-after-keywords": [
-      2,
-      "always"
-    ],
+    "keyword-spacing": 2,
     "space-in-parens": [
       2,
       "never"
@@ -98,7 +95,6 @@
     "eqeqeq": [2, "allow-null"],
     "no-alert": 2,
     "no-caller": 2,
-    "no-empty-label": 2,
     "no-extend-native": 2,
     "no-extra-bind": 2,
     "no-fallthrough": 2,
@@ -154,8 +150,7 @@
         "after": true
       }
     ],
-    "semi": 2,
-    "space-return-throw-case": 2
+    "semi": 2
   },
   "env": {
     "node": true,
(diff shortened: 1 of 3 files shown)


dcchuck:
Happy to hear your thoughts, I believe that's a relic of previous settings?

Dwaynekj:
@joneit What are you thinking for merge settings?

joneit:
I'm not religious about this. *Ideally,* a PR should contain a single commit for a single task. Multiple tasks should be split out into multiple PRs. However, there are problems with this ideal: 
1. Sometimes a set of tasks are based one on the other and if submitted as separate PRs they introduce ordering dependency and very often delays which works against agile.
2. Often, a task is poorly defined, and turns into several tasks along the way, especially for unforeseen problems.

There are always pros and cons:
### Pros
* Squash & merge gets rid of noise
### Cons
* Squash & merge looses dev history
* Squash & merge creates an entirely new commit

I for one go out of my way to keep my commits to the minimum by squashing as I go. I would still want to submit a PR with multiple commits, however, when those commits represent logical and concise views of the progress of the development, especially when the PR ended up being one of those comprised of several subtasks. For complex PRs, this can really aid reviewers to see relevant logic changes together.

I realize that some people use commits just to "save their work" and end up with 20 or 30 commits. In such cases, we could ask them to squash it and resubmit it.

Regarding GitHub creating new commit, this means that your original commit SHAs are lost. After pulling from develop to get the new commit, you now have to delete both your local branch and your remote tracking branch.

All of this contributes to neatness, but there is a loss of historical information, such as names, dates, and history.

Until things actually get out of hand, I would vote to not do squash & merge.

Dwaynekj:
Hmm ok. We can certainly add it back later

joneit:
@Dwaynekj Thanks for approving. I went ahead and merged this.

We want approvals from both teams in future, but there will be times, like this one, where 3 people (1 from each team + PR author who cannot approve his own PR) is overkill or not available.

Issue #641, opened 2017-12-08
Bad refs to rowColumnIndex and treeColumnIndex


joneit:
[These lines](https://github.com/fin-hypergrid/core/blob/develop/src/renderer/index.js#L661-L667) , the context for`rowColumnIndex` and `treeColumnIndex` should be `this.grid.behavior` (not `this.grid`).

This resulted in `getColumnWidth()` always returning the default column width instead of the actual column width (not obvious when the actual column with _was_ the default).

This is a problem in develop, introduced in #598: _fix keyboard navigation_; it is not a problem in master.

Issue #634, opened 2017-11-14
data model options


joneit:
@Dwaynekj Hey DJ, While refining/updating issue #614 (v3 prep), I came across your change to the calling signature of the data source constructor, which I’m sure you recall was in support of group view's tree column.

You added two parameters to the constructor. While pertinent to the current “standard” datamodel, it is meaningless to a generic “black box” datamodel, which should not need to support those parameters.

I propose changing it to a generic options object. I'll post a PR.

cc:@dcchuck

Dwaynekj:
Ok, I'll await the PR

joneit:
@Dwaynekj Change of plan: I'm going to submit this today as a stand-alone PR to facilitate testing and review. (Moving away from large PRs.)

PR #643, opened 2017-12-11
Remove treeColumnIndex/rowColumnIndex from DataSourceOrigin

joneit:
See issue #634 except this PR removes `treeColumnIndex` and `rowColumnIndex` from the data source entirely (without implementing an `options` object to hold them, as originally recommended in the issue).

Also addresses issue #641.

### DIFF
--- a/src/behaviors/Behavior.js
+++ b/src/behaviors/Behavior.js
@@ -131,21 +131,25 @@ var Behavior = Base.extend('Behavior', {
         return this.grid.renderer.visibleRows.length;
     },
 
-    get treeColumnIndex() {
-        return -1;
-    },
-
-    get rowColumnIndex() {
-        return -2;
-    },
-
     get leftMostColIndex() {
         return this.grid.properties.showRowNumbers ? this.rowColumnIndex : (this.hasTreeColumn() ? this.treeColumnIndex : 0);
     },
 
     clearColumns: function() {
-        var tc = this.treeColumnIndex;
-        var rc = this.rowColumnIndex;
+        var schema = this.dataModel.schema,
+            tc = this.treeColumnIndex,
+            rc = this.rowColumnIndex;
+
+        schema[tc] = schema[tc] || {
+            name: 'Tree',
+            header: 'Tree'
+        };
+
+        schema[rc] = schema[rc] || {
+            name: '',
+            header: ''
+        };
+
         /**
          * @type {Column[]}
          * @memberOf Behavior#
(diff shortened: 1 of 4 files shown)


joneit:
@Dwaynekj The new build works fine but the standard demo does not use tree column... Can you tell me how to test tree column? (or can you test it?)

Dwaynekj:
The Tree,Agg and Group View starter demo should use the tree-column.  How will we support in in the blackbox sense?

joneit:
@Dwaynekj you'll have to fill me in on what exactly you had in mind here. Perhaps when we get together … ?

I don't think the question should be "what do we have to support [for Barclays]?" but "what is a good flexible extensible interface for supporting ________?" (fill in the blank: "synthetic columns" or something).

But as a starting point for discussion:

### Originally we had

index|column|notes
:-:|---|---
`-1`|row numbers|1st column, always visible
`-2`|tree|2nd column, always visible
`0`|data|first data column, visible when not horizontally scrolled out of view

### Your change
For some reason you wanted the lexical order of the keys to reflect the display order of the columns:

index|column|notes
:-:|---|---
`-2`|row numbers|1st column, always visible
`-1`|tree|2nd column, always visible
`0`|data|first data column, visible when not horizontally scrolled out of view

### What I thought you had in mind
When no tree column, row number column would be `-1` instead of `-2` — which is why you pass them to the data source (because they're dynamic)

### What I see however
Looks to me like:
1. row number column is _always_ `-2` and if tree column not defined, we just skip it ([here](https://github.com/fin-hypergrid/core/blob/develop/src/renderer/index.js#L17-L19) and [here](https://github.com/fin-hypergrid/core/blob/develop/src/renderer/index.js#L32-L34)).
2. both columns are always defined regardless of usage because [the getters](https://github.com/fin-hypergrid/core/blob/develop/src/behaviors/Behavior.js#L134-L140) are hardcoded to these values.
3. and furthermore, it's hard-coded at least [in `DataSourceAggregator.js`](https://github.com/fin-hypergrid/old-addons/blob/master/aggregations/dataSources/DataSourceAggregator.js#L27).

### So, I'm confused about what's going on here
And I'm thinking I can simplify this whole thing _once and for all_ if we can clearly define what the actual needs are. Let's talk!

joneit:
@Dwaynekj I've already rebased 3.0 to these commits, so I'm going to merge this now. If needed, we can tweak later to keep client expectations from breaking.

PR #646, opened 2017-12-11
fixed canvas key event map typo

stevewirts:


joneit:
@stevewirts thanks for resubmitting (*frown* shouldn't have been necessary to make a new PR)

Issue #647, opened 2017-12-11
Row Props


joneit:
Let's discuss row props.

The only row prop we currently support _for individual rows_ is height.

We also have `grid.properties.rowProperties` which is an array applied modulo to the grid; each element of the array is a properties collection. This is for easy row striping.

What we *don't* have is a general individual row properties object. My concern is performance degradation but the concern may be overblown.

The code is easy enough. I'm going to work up a PR for this and we can put it through some excercises.

joneit:
General row prop support added in [2.1.0](https://github.com/fin-hypergrid/core/wiki/Release-Notes:-v2.1.0#row-properties).

PR #648, opened 2017-12-11
Implement true row props

joneit:
See issue #647. ~This PR is tentative based on performance impact.~ _Code has now gone through several rounds of testing/tweaking for performance. Performance checks out; please proceed with review._

### DIFF
--- a/demo/row-props.html
+++ b/demo/row-props.html
@@ -139,35 +139,35 @@ Renders all cells over consolidated row backgrounds, if any. Unsuitable as a cli
     </div>
     <div class="radio-group">
         Grid row properties:
-        <label title="Removes rowProperties array from grid.properties entirely.)">
-            <input type="radio" name="striped-rows" onclick="rowProps(null)" checked>
+        <label title="Removes `stripes` array from grid.properties entirely.)">
+            <input type="radio" name="striped-rows" onclick="stripes(null)" checked>
             none
         </label>
-        <label title="Adds a rowProperties array to grid.properties that defines 3 white rows alternating with 3 light green rows. (Row rects are NOT drawn for those first 3 row rects when and only when the grid background is also white.)">
-            <input type="radio" name="striped-rows" onclick="rowProps({ backgroundColor: 'white' })">
+        <label title="Adds a `stripes` array to grid.properties that defines 3 white rows alternating with 3 light blue rows. (Row rects are NOT drawn for those first 3 row rects when and only when the grid background is also white.)">
+            <input type="radio" name="striped-rows" onclick="stripes({ backgroundColor: 'white' })">
             3 white + 3 blue
         </label>
-        <label title="Adds a rowProperties array to grid.properties that defines 3 grid background rows alternating with 3 light green rows. (Row rects are NOT drawn for those first 3 undefined rows.)">
-            <input type="radio" name="striped-rows" onclick="rowProps(undefined)">
+        <label title="Adds a `stripes` array to grid.properties that defines 3 grid background rows alternating with 3 light blue rows. (Row rects are NOT drawn for those first 3 undefined rows.)">
+            <input type="radio" name="striped-rows" onclick="stripes(undefined)">
             3 undefined + 3 blue
         </label>
     </div>
-    <label title="Column properties defines a light blue background color BUT inherit's grid's rowProperties array.">
+    <label title="Column properties defines a light blue background color BUT inherit's grid's `stripes` array.">
         Headers:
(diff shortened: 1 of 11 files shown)


nmichaud:
What's the difference between this and #640?

joneit:
@nmichaud Whoops forgot to push the 2nd commit; it's pushed now

joneit:
Rebased ("2nd commit" is now the only commit).

joneit:
As a side note:

It would be nice if Hypergrid had a property to specify the properties hierarchy because I can see how some apps would prefer a different order, for example:
> grid ← row ← column ← cell

Unfortunately, this is not possible in a performant way due to the current design. However it may be possible in a future version if we can address the design issue.

joneit:
I did some performance profiles and performance is not measurably impacted, which is great.

Deprecated `grid.properties.rowProperties` (with a warning) in favor of a better name `stripes` because we now have true row properties and this was now confusing.

Implemented following new properties, defaulted to `true`, which if set to `false` will disable rendering the related properties, significantly improving performance.
* `renderColumnProps`
* `renderStripes`
* `renderRowProps`
* `renderCellProps`

I also isolated the logic for layering properties during rendering into a new method `grid.renderer.configurator` so that developer can override it with different or faster logic.

joneit:
@Dwaynekj @dcchuck @nmichaud @neilslinger @texodus @stevewirts Success! This PR passed performance tests so it's definitely a go, so _**please review.**_

Note that row properties api is simpler and shorter cell properties api — which we could consider simplifying as well.

Dwaynekj:
@joneit what Browsers and OS did you test? Chrome 40 is in that list?

joneit:
@Dwaynekj Good point. I'll test in Chrome 40. But this can still go because of the props and the overridable method:
* If impactful to Chrome 40l, I'll try setting `grid.properties.renderRowProps = false` to skip the logic ([the line]() I called out in previous comment).
   * If better, we can consider making that the default? ...although upgrading to v3.0.0 is going to require some code tweaks anyway and setting this prop to `false` but I hesitate doing that for the relatively few Chrome 40 users; rather it could be a "recommended Chrome 40 tweak."
* If still impactful, we can supply a simpler `configurator` akin to the original logic (by eliding that line entirely).
   * Again, we could consider making that the default, but again that setting could be a recommended Chrome 40 tweak.

stevewirts:
I WILL be there at 330 today

joneit:
## Layering prop classes in any order

I've re-pushed this PR with new logic to select prop classes to render and in which order. Overcame some performance-related issues to do this. 

### Performance checks out!
I did some performance profiles and performance is not measurably impacted, which is great.

There is still one case where performance takes a hit (when column props are omitted; see below). 

### Renamed property: `rowProperties` → `stripes`
Deprecated `grid.properties.rowProperties` (with a warning) in favor of a better name `stripes` because we now have true row properties and this was now confusing.

### New property: `cellEvent.prototype.rowProperties`
This is a new getter/setter ([example](https://github.com/fin-hypergrid/core/pull/648/files#diff-5a077d9e9d2adeca2bd0274d0a40ed5dR1242)).

The logic layers the property classes in the order defined in `Hypergrid.properties` (aka `require('defaults.js')`):
```js
propClassLayers: [ 'columns',  'stripes',  'rows',  'cells' ]
```
In addition to the above grid default, the tree and row handle columns both override the grid property ([actual code](https://github.com/fin-hypergrid/core/pull/648/files#diff-efa8c1cff4b1c772c558c7576aa52fd0R174)):
```js
treeColumnProperties.propClassLayers = rowColumnProperties.propClassLayers = ['columns'];
```
The above only renders the column props, thus preventing row stripes from rendering in those columns.

### Example
So the default is to override column props with row props. To override this default to render column props on top of row props:
```js
grid.properties.propClassLayers = ['rows', 'columns'];
```

### New overridable method: `grid.renderer.assignProps`
The layering logic is isolated in a [new method](https://github.com/fin-hypergrid/core/pull/648/files#diff-5a077d9e9d2adeca2bd0274d0a40ed5dR1239) which developers can override with alternative or faster logic.

joneit:
Rebased to develop post merge of #661.

joneit:
@dcchuck do give me a call if you'd like a demo and/or walk-thru!

joneit:
@nmichaud @dcchuck @Dwaynekj @neilslinger Code has now gone through several rounds of testing/tweaking for performance. Performance checks out. We need at least one review (better would be two) from you guys.... thanks!

joneit:
Rebased to develop.

joneit:
Rebased to `develop` required after merging #666, #674, and #676.

Issue #617, opened 2017-07-21
IE 11 Rendering Quirks


Dwaynekj:
* While hovering over the grid, the cursor has a ‘text’ style – it’s a pointer in other browsers
* On focusing into the grid, by selecting a cell, a text cursor appears at the left top corner (see screenshot) and remains there until one of the cells is edited

![image](https://user-images.githubusercontent.com/1011448/28472055-2d6d6d34-6e0d-11e7-9110-19fcc024ed64.png)

PR #650, opened 2017-12-12
FIx issue in IE with I-beam cursor

nmichaud:
Not sure why content editable was ever set on a canvas element, but in IE it causes the cursor to always be the I-beam shape. This fixes the first part of #617


joneit:

PR #651, opened 2017-12-12
Fire row selection event when auto selecting rows

nmichaud:
Original code was missing the emission of event.


joneit:

PR #652, opened 2017-12-12
Cell bounds should account for whether grid lines are rendered

nmichaud:


nmichaud:
@joneit this should account for the new changes in #664

joneit:
@nmichaud 
> this should be updated to account for the new changes in #664

Right.... I'm closing the PR and creating a new one based on #664.

joneit:
Opening this back up for discussion...

joneit:
@nmichaud I don't see what problem this is solving.

The only difference I see in rendering is when grid lines are off (`gridLinesV=gridLinesH=false`):
* The cell rects become slightly larger and the round-off from centering the text is affected, making the text appear to move slightly.
* No gaps appear between cells when selecting multiple cells (or on the hover highlights).

This is exactly the same effect you would get from setting `lineWidth=0` — except that you're setting the vertical line width (your `lineWidthX`) to `1` instead of `0` so your text only moves down slightly while previously it would have moved down and right.... I suspect the `1` had something to do with the way you were doing the fixed/scrolling boundary rule line stuff, though it is unclear to me what exactly.

The current implementation is actually more flexible, and changing it would change the behaviorof existing apps with regards to `gridLinesH` and `gridLinesV`. The only missing piece being that we do not have separate vertical and horizontal line widths, so you cannot currently eliminate the gaps between rows while keeping the gaps between columns or vice versa.

I'm not clear what you were looking for here:
* Was this something that was necessary for your boundary rule line logic that perhaps is no longer necessary?
* Was the gaps between cells of interest to you?

If the latter, implementing separate line widths would give you that.

joneit:
Per our discussion, am closing and will submit a new PR for independent horizontal/vertical line width properties.

PR #653, opened 2017-12-12
Don't do anything if there's no data

nmichaud:
If the grid has no data, this function starts throwing spurious console errors as the user mouses over the empty grid.


joneit:

Issue #660, opened 2017-12-21
Defer "changed" methods


joneit:
The following should be able to be called multiple times, similar to `repaint()`, while deferring and collapsing all calls until the end of the thread:

* `behaviorChanged`
Does one of the following:
* `behaviorShapeChanged`
Recalculates scrollbars, followed by...
* `behaviorStateChanged`
Recalculate rects of visible cells, followed by a call to `repaint()`

Right now the two recalc operations are performed on every call. For this reason we have always tried only to make such calls near end of thread, sometimes omitting calls where logically needed to avoid multiple recalcs but sometimes no doubt incurring multiple recalcs needlessly

PR #661, opened 2017-12-21
Defer "xxxxChanged()" methods

joneit:
See Issue #660.

I tried moving the "changed"-related code to a separate file but it was only 60 lines so I undid it.

I did however move all scrolling-related members of `Hypergrid.prototype` to a separate file and then mixed it in.

### DIFF
--- a/src/lib/modules.js
+++ b/src/lib/modules.js
@@ -1,5 +1,16 @@
 'use strict';
 
+/*
+ * This module is the namespace of loaded external modules known to `Hypergrid.require`,
+ * which may include loaded application modules, datasource modules, and plug-in modules.
+ *
+ * The pre-loaded external modules listed below can conveniently be overridden by the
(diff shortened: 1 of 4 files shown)


joneit:
@Dwaynekj @dcchuck @nmichaud @neilslinger @texodus @stevewirts I'd like to get this merged to develop. The 3.0 PR is built on top of it and until it's merged, its commit is showing up in the 3.0 diff.

src/lib/modules.js
```diff
@@ -0,0 +1,5 @@
+'use strict';
+
+module.exports = {
+    Scrollbar: require('finbars')
```
dcchuck: Is the idea here to eventually move all imports to this file? Just wondering why this abstraction

joneit:
@dcchuck Oh nevermind I see your approval.

PR #663, opened 2017-12-31
Use fin-hypegrid-event-logger in demo

joneit:
[`fin-hypegrid-event-logger`](https://github.com/fin-hypergrid/event-logger) is a new plugin that uses my new [`starlog`](https://www.npmjs.com/package/starlog) event logger, something I was playing with over break.

The logging output (when `demo.vent = true`) is basically the same as it was.

This PR:
* Loads the plugin and connects it to `demo.vent`.
* Moves `if (vent) { ... }` custom logging code to the plug-in ([here](https://github.com/fin-hypergrid/event-logger/blob/master/custom-listeners.js)).
* Changes selection array event properties (`.rows`, `.columns`, and `.selections`) to getters so they are not executed unless actually referenced.

### DIFF
--- a/src/Hypergrid.js
+++ b/src/Hypergrid.js
@@ -71,6 +71,8 @@ var EDGE_STYLES = ['top', 'bottom', 'left', 'right'],
  */
 var Hypergrid = Base.extend('Hypergrid', {
     initialize: function(container, options) {
+        this.selectionInitialize();
+
         //Optional container argument
         if (!(typeof container === 'string') && !(container instanceof HTMLElement)) {
             options = container;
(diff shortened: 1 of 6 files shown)


joneit:
@nmichaud Good point. I rolled it back here and will submit a new PR for lazy event props.

joneit:
Reviewers @nmichaud @dcchuck @Dwaynekj do feel free to comment on /event-logger repo (via new issue](https://github.com/fin-hypergrid/event-logger/issues)).

src/Hypergrid.js
```diff
@@ -89,7 +91,7 @@ var Hypergrid = Base.extend('Hypergrid', {
         );
 
         // Install shared plug-ins (those with a `preinstall` method)
-        Object.getPrototypeOf(this).installPlugins(options.plugins);
+        Hypergrid.prototype.installPlugins(options.plugins);
```
nmichaud: This change seems unrelated to the PR?

joneit: that is true.... it is unimportant & trivial & need to put it somewhere

joneit: @nmichaud aside from this are you ready to approve?

joneit:
btw, lazy event props were moved to new PR #671

nmichaud:
Yeah looks good to me

PR #656, opened 2017-12-12
Draw a visual indication for fixed columns/rows when edge visible

nmichaud:
This feature draws a slight visual indicator of the fixed rows/columns (basically a ~1~ _`lineWidth`_-pixel line if grid lines are not shown or a ~slightly thicker~ _`(lineWidth + 1)`_-pixel line if they are~n't~).

PR #664, opened 2018-01-03
flexible rule lines between fixed & scrolling rows & columns

joneit:
More flexible alternative to _Draw a visual indication for fixed columns/rows when edge visible_ (#656) which had linter and other errors.

This version adjusts the grid render to _make room for_ the rule lines between the fixed and scrolling sections, which can now be of _any width_ (new grid property `fixedLineWidth`) — whereas the original version had no property, did not make room for the extra thick line, just _squeezing in_ an extra column (and/or row) of pixels (which is all there was room for).

As this new design allows us to have any width, I set the default to 3 rather than 2 which IMHO works better.

It also has an options for:
* Define `fixedEdgeWidth` for "double" rule line (of any scale) (defaults to `undefined`)
* Define `fixedLineColor` for specific color for this rule line (defaults to `undefined`)

`fixedEdgeWidth` works together with `fixedLineWidth` to define "double" rules where the rule line is rendered as two `fixedEdgeWidth`-wide colored pixel row(s) separated by whatever width remains, specifically, `(fixedLineWidth - fixedEdgeWidth - fixedEdgeWidth)` background row(s) of pixels.

For example:
* `fixedLineWidth = 3; fixedEdgeWidth = 1` — rendered as a double rule line consisting of two 1-pixel rows separated by (3 - 1 - 1 = 1) empty row
* `fixedLineWidth = 7; fixedEdgeWidth = 2` — rendered as a double rule line consisting of two 2-pixel rows separated by (3 - 2 - 2 = 3) empty rows

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -25,7 +25,7 @@
     </div>
 
     <!-- recommended height: multiple of default row height, plus horizontal scroll bar height -->
-    <div id="json-example" style="margin-top:12px; margin-left:30px; height:402px"></div>
+    <div id="json-example" style="margin-top:12px; margin-left:30px; height:404px"></div>
 
 </body>
 </html>
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,9505 @@
+{
+  "name": "fin-hypergrid",
+  "version": "2.0.2",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "Base64": {
+      "version": "0.2.1",
+      "resolved": "https://registry.npmjs.org/Base64/-/Base64-0.2.1.tgz",
+      "integrity": "sha1-ujpCMHCOGGcFBl5mur3Uw1z2ACg=",
+      "dev": true
+    },
+    "JSONStream": {
+      "version": "0.7.4",
+      "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-0.7.4.tgz",
+      "integrity": "sha1-c0KQ5BUR7qfCz+FR+/mlY6l7l4Y=",
+      "dev": true,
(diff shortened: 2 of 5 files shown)


joneit:
Removed support for `gridBorderTop`, _etc.,_ from `paintGridlines`.

180f376 makes them into _dynamic properties_ which set the CSS borders of the `<canvas>` element.

This is a change because:
1. Previously the borders were painted; now they are CSS.
2. Previously they were inside; now they are outside.
3. Previously the right border was painted along the right edge of the last visible column; now it is rendered at the right edge of the canvas which may be much wider than the rendered grid.

So _why_ the change? Because the grid was not making room for the frame and the borders were painting over the content. Tended to be OK if the grid frame was only 1 pixel wide. However, 1 pixel is rather too thin for a grid frame anyway, so not too useful and this feature did not see much use. We could make room, but it's "wasted" pixels because CSS does this beautifully anyway. **A better approach, which this change reinforces, is to let CSS render the borders at any width _outside_ the canvas.**

Another question might be, so _why_ make it part of Hypergrid properties at all? Why not just deprecate it and let app developer set borders. The answer is that it can still properly be thought of as a grid property, persisting with other grid properties to reconstruct the look of the grid next time in.

The CSS borders could of course be rendered inside the box, by setting `grid.canvas.canvas.style.boxSizing = 'border-box'`.  As noted in [comment](https://github.com/fin-hypergrid/core/pull/664/files#diff-3c2d012a733d2ddb11fb77104023f7e9R469), however, this squashes the canvas slightly, rendering the text blurry … so _not recommended._

joneit:
Rebased to develop after merge of #665 + #669.

joneit:

PR #666, opened 2018-01-03
Feature chain property

joneit:
A property-based alternative to `behavior.features`.

Now the feature chain can be specified by `grid.properties.features` which is an array of registered feature names.

The advantage of making it a serializable property is that the feature chain can now be persisted along with other grid properties.

Caveat: Feature chain is delicate and dependent on the order of features. This approach is useful for omitting or substituting features, so long as the basic order is retained. Rearrange the order at your own risk!

Overriding `Behavior.prototype.features`, an array of feature constructors, is still supported and will get priority. (Override can take form of assigning to prototype before grid instantiation or  subclassing behavior with an override in subclass's prototype, and handing that subclass to `Hypergrid` constructor, _etc._)

### DIFF
--- a/demo/js/demo/celleditors.js
+++ b/demo/js/demo/celleditors.js
@@ -6,7 +6,7 @@ module.exports = function(demo, grid) {
 
     var idx = grid.behavior.columnEnum;
 
-    var CellEditor = grid.cellEditors.get('celleditor');
+    var CellEditor = grid.cellEditors.BaseClass;
     var Textfield = grid.cellEditors.get('textfield');
 
     var ColorText = Textfield.extend('colorText', {
--- a/demo/js/demo/cellrenderers.js
+++ b/demo/js/demo/cellrenderers.js
@@ -188,7 +188,7 @@ module.exports = function(demo, grid) {
 
 
     //Extend HyperGrid's base Renderer
-    var sparkStarRatingRenderer = grid.cellRenderers.get('emptycell').constructor.extend({
+    var sparkStarRatingRenderer = grid.cellRenderers.BaseClass.extend({
         paint: paintSparkRating
     });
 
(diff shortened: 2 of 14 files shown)


nmichaud:
Not sure I like the idea of having both ways of specifying feature chains. Seems it could lead to some subtle bugs.

joneit:
@nmichaud 
>Not sure I like the idea of having both ways of specifying feature chains. Seems it could lead to some subtle bugs.

Well it's true that the old way did not require features to be "registered"... although the only reason we're registering them (so far) is to make them accessible by name....

The main reason I kept the old way functional is for backwards compatibility. If we remove it, it would break existing apps. Not sure there are any that play with the feature chain. Maybe yours was the only one?

A compromise would be to issue a deprecation warning (and terminate support at a later date). What do you think of that?

joneit:
@nmichaud 
> Sure, lets go with the deprecation approach.

I added a deprecation warning (see sample below).

**However, on thinking about this, we should support both approaches** because they serve different audiences:
* When an app always uses the same feature chain for all its grid instances, `grid.behavior.features` (array of feature constructors) is preferred.
* Only when the feature chain may vary per grid within the same app is `grid.properties.features` (serializable array of feature names) really useful. Otherwise its a bit of a nuisance, frankly.

**Conclusion:** Let's _not_ deprecate `grid.behavior.features`. The current code supports both approaches (giving priority to `grid.properties.features`, when defined).

Sample console warning:
```
Behavior.js:1453 `grid.behavior.features` (array of feature constructors) has been deprecated
as of version 2.1.0 in favor of `grid.properties.features` (array of feature names). Remove
`features` array from your behavior and add `features` property to your grid state object
 (or Hypergrid.defaults), e.g.:

Hypergrid.defaults.features = [
	'Filters',
	'CellSelection',
	'KeyPaging',
	'ColumnResizing',
	'RowSelection',
	'ColumnSelection',
	'ColumnMoving',
	'ColumnSorting',
	'CellClick',
	'CellEditing',
	'OnHover',
];
```
And the following is added when a listed feature is not currently registered:
```
The following features are unregistered and will need to be registered prior to
behavior instantiation:

Features.add('yada');
Features.add('yada-yada');
```

joneit:
@nmichaud Please finish review thx.

PR #668, opened 2018-01-04
SHIFT can be the first key in the list

nmichaud:
Quick fix in logic to check whether keys contains the Shift key. If Shift is the first element of the array indexOf returns 0.

### DIFF
--- a/src/features/RowSelection.js
+++ b/src/features/RowSelection.js
@@ -202,7 +202,7 @@ var RowSelection = Feature.extend('RowSelection', {
         if (!grid.abortEditing()) { return; }
 
         var mouseY = grid.getMouseDown().y,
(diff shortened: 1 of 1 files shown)


src/features/RowSelection.js
```diff
@@ -202,7 +202,7 @@ var RowSelection = Feature.extend('RowSelection', {
         if (!grid.abortEditing()) { return; }
 
         var mouseY = grid.getMouseDown().y,
-            hasSHIFT = keys.indexOf('SHIFT') > 0;
+            hasSHIFT = keys.indexOf('SHIFT') !== -1;
```
joneit: i like `>= 0` but same diff! approving...

joneit:

PR #654, opened 2017-12-12
Implement column sorting using single click

nmichaud:
This might be a bit more controversial, but our users often like prefer single clicking on the headers to sort.

PR #669, opened 2018-01-04
Removed double-click delay

joneit:
> NOTE: This new PR replaces _Implement column sorting using single click_ (#654), which is now closed.

#### Defunct option `doubleClickDelay`

This PR removes the `doubleClickDelay` property so single-clicks are more responsive. The double-click delay is no longer defined in properties. Trying to set it results in a warning that it is now ineffective and the double-click delay is now system-dependent (_e.g.,_ in mouse control panel).

For the record, the defunct property defaulted to 325ms.

The naming of this property parallels the existing `editOnDoubleClick` property. Cell editing now respects cell's or column's property setting rather than relying exclusively on grid property setting.

#### New option: `sourtOnDoubleClick`
Also added an option to allow sort click to be single-click or double-click (previously was hard-coded to double-click). The new property `sortOnDoubleClick` defaults to `true`; change this to `false` for single-click.

When `false` (sort is single-click), both sorting and column selection toggling occur on the same click.

When `true` (sort is double-click), we have the very situation warned against on [the web](https://api.jquery.com/dblclick):

> It is inadvisable to bind handlers to both the click and dblclick events for the same element. ...

In this case (and only in this particular case) what we do is delay before responding to single click (for column selection). This is the only double-click delay remaining in the code base.

We cannot unfortunately read the system double-click delay so I have hard-coded the above delay to 300ms.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fin-hypergrid",
-  "version": "2.0.2",
+  "version": "2.1.0",
   "description": "Canvas-based high-performance spreadsheet",
   "repository": {
     "type": "git",
--- a/src/defaults.js
+++ b/src/defaults.js
@@ -900,6 +900,8 @@ var defaults = {
     editable: true,
 
     /**
+     * Edit cell on double-click rather than single-click.
+     *
      * @default
      * @type {boolean}
      * @memberOf module:defaults
@@ -907,11 +909,16 @@ var defaults = {
     editOnDoubleClick: true,
 
     /**
+     * Sort column on double-click rather than single-click.
+     *
+     * Used by:
+     * * feature/ColumnSorting.js to decide which event to respond to (if any, see `unsortabe`).
+     * * feature/ColumnSelection.js to decide whether or not to wait for double-click.
      * @default
-     * @type {number}
+     * @type {boolean}
      * @memberOf module:defaults
      */
-    doubleClickDelay: 325,
+    sortOnDoubleClick: true,
 
     /**
      * Grid-level property.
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fin-hypergrid",
-  "version": "2.0.2",
+  "version": "2.1.0",
```
nmichaud: Shouldn't we change this in a separate commit right before we cut a version?

joneit: @nmichaud sure, i'll pull it out

src/defaults.js
```diff
@@ -942,6 +940,55 @@ var defaults = {
...
+    /**
+     * **This is a standard property definition for sort plug-in use.
+     * It is not referenced in core.**
+     *
+     * The maximum number of columns that may participate in a multi-column sort (via ctrl-click headers).
+     * @default
+     * @type {number}
+     * @memberOf module:defaults
+     */
+    maxSortColumns : 3,
```
nmichaud: Is there a reason to have a limit?

joneit: @nmichaud 
> Is there a reason to have a limit?

I think the reason was the extra crunching involved. Implementations are free to ignore the property. Applications are free to set it very high. We could set the default higher (like 5 or 6)? I wouldn't eliminate it entirely; t's a handy property to keep around.

What do you think?

nmichaud: Well if implementations are free to ignore it, i'm not sure how useful it is.

joneit: We have no way of enforcing that applications respect the property values. If they do, it becomes "easier" (not _guaranteed_) that one sorting plug-in can be swapped out for another. I think it is better to have it than not to have it. And on a practical level, apps can override it if it is a problem. We can also, as I said make the default higher. What do you think?

joneit:
(Merging with only @nmichaud approval, per @dcchuck.)

PR #671, opened 2018-01-04
Lazify event selection props

joneit:
These props are costly to run so should not run unless actually needed.

Even though these were only invoked rarely (on specific user interactions), they could be _very_ costly if a lot of rows/columns are involved, in terms of memory and execution time.


joneit:
I'm going to merge this now — without waiting for a 2nd approval — because it contains new method `selectionInitialize`, the call to which was in #663, _already merged._

(Obviously neglected to retest after teasing out of #663. Whoops.)

PR #674, opened 2018-01-05
Distinct vertical & horizontal line widths

joneit:
**\[Supercedes #652, now closed without merge.]**

### Distinct props

Although we had distinct properties, `gridLinesH` and `gridLinesV`, for enabling grid lines on the 2 axes, the other gridline properties, `lineColor` and `lineWidth`, pertained to both grid line axes simultaneously. **This PR creates distinct color and width properties.**

Implemented as a new _complex property_ (a property that is an object rather than a scalar), which gathers the 3 props (width, color, enabled) for each of 2 axes (vertical, horizontal) = 6 props all together into a single object [as defined in defaults.js](https://github.com/joneit/core/blob/935384d10c9874a874d51e30a700e17bba95a051/src/defaults.js#L449-L473)):
```js
    . . .
    gridLines: {
        horizontal: {
            enabled: true,
            width: 1,
            color: 'rgb(199, 199, 199)'
        },
        vertical: {
            enabled: true,
            width: 1,
            color: 'rgb(199, 199, 199)'
        }
    }
    . . .
```

The original properties have been retained but they are now dynamic properties that operate on the new properties as follows:

Original<br>Property | New<br>Property | Setting<br>Original<br>sets… | Getting<br>Original<br>gets…
---|---|:-:|:-:
`gridLinesH` | `gridLines.horizontal.enabled` | same | same
`gridLinesV` | `gridLines.vertical.enabled` | same | same
`lineColor` | `gridLines.horizontal.color`<br>`gridLines.vertical.color` | both | horizontal
`lineWidth` | `gridLines.horizontal.width`<br>`gridLines.vertical.width` | both | horizontal

The original props are now non-enumerable so they are ignored by `grid.saveState()`.
The new `gridLines` prop they operate on is output by` saveState` instead.

### Fixed complex props issue

Fixed a pre-existing issue with complex props. Changes inside such objects in the theme or defaults layers were affecting all other grid instances. Solution was to deep clone all default props of object type (excluding dynamic props).

### DIFF
--- a/src/Hypergrid.js
+++ b/src/Hypergrid.js
@@ -268,6 +268,17 @@ var Hypergrid = Base.extend('Hypergrid', {
             grid: { value: this },
             var: { value: new Var() }
         });
+
+        // Deep clone all default props of object type, excluding dynamic props, so changes inside
+        // such objects won't affect theme or defaults layers which may be shared by other instances.
+        Object.keys(defaults)
+            .filter(function(key) { return !dynamicPropertyDescriptors[key]; })
+            .forEach(function(key) {
+                var value = Object.getOwnPropertyDescriptor(defaults, key).value; // do not invoke getters!
+                if (typeof value === 'object') {
+                    this[key] = deepClone(value);
+                }
+            }, this.properties);
     },
 
     /**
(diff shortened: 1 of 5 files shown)


joneit:
Rebased to develop.

PR #676, opened 2018-01-09
Hide row checkboxes & row numbers independently

joneit:
To hide one feature or the other, reset the new properties `rowHeaderFeatures.checkboxes` and `rowHeaderFeatures.numbers`. If both are reset, the entire column is hidden.

Setting the legacy property `showRowNumbers` still hides and shows the entire column by setting both of the above. Getting the legacy property returns `true` if either of the above are set.

> NOTE: First commit below, 0e7eff6, belongs to PR #674; please review that PR first; don't comment here on edits coming from that commit.

### DIFF
--- a/demo/js/demo/index.js
+++ b/demo/js/demo/index.js
@@ -37,7 +37,6 @@ window.onload = function() {
         grid = window.grid = window.g = new Hypergrid('div#json-example', gridOptions),
         behavior = window.b = grid.behavior,
         dataModel = window.m = behavior.dataModel,
-        initial = true,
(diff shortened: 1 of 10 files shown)


joneit:
@nmichaud I pushed another commit (then force-repushed it) to fix a [bad edit](https://github.com/fin-hypergrid/core/pull/676/commits/871762627db8138c1f8f541d26b14a8329ab0829#diff-ca8fe17906e25a4ffd2911c969658fafL201). **Please approve again.**

PR #685, opened 2018-01-23
Post-2.1.0-release fixes (2.1.1)

joneit:


joneit:
Hey @Dwaynekj - I added a commit to this to revert the breaking change so we can publish to npm without further concern.

src/lib/DataSourceOrigin.js
```diff
@@ -79,7 +82,7 @@ var DataSourceOrigin = DataSourceBase.extend('DataSourceOrigin',  {
                 schema[i] = { name: fields[i] };
             }
         }
-
+        schema = setInternalColSchema.call(this, schema);
```
Dwaynekj: What does this accomplish?

joneit: Just reverting the breaking change back to [__your__ code](https://github.com/fin-hypergrid/core/commit/c540b45ace615d757147861a48c26460493455b1#diff-4a3b59fc30ab47d3eee9791d7bc35230R335)!

Dwaynekj: ok

joneit:
@Dwaynekj Pushed another commit to resolve an issue I just fell across.

PR #706, opened 2018-03-16
v2.1.6

joneit:
* Fixed COPY (ctrl-C, command-C) regression (copy select data command)
* Fixed single-column-after-fixed-column(s) render issue
* For restoring row selection(s) across data transformations (`reindex` calls):
   * Conditional upon new `restoreRowSelections` property (was formerly conditional upon `checkboxOnlyRowSelections` property)
   * Now also restores column selection(s), conditional upon new `restoreColumnSelections` property
* `Behavior.prototype.getActiveColumnIndex` now accepts a column name (string) (in addition to a data row index or column object)

### DIFF
--- a/src/Hypergrid.js
+++ b/src/Hypergrid.js
@@ -19,8 +19,8 @@ var SelectionModel = require('./lib/SelectionModel');
 var Localization = require('./lib/Localization');
 var Behavior = require('./behaviors/Behavior');
 var behaviorJSON = require('./behaviors/JSON');
-var CellRenderers = require('./cellRenderers');
-var CellEditors = require('./cellEditors');
+var cellRenderers = require('./cellRenderers');
+var cellEditors = require('./cellEditors');
 
 var EDGE_STYLES = ['top', 'bottom', 'left', 'right'],
     RECT_STYLES = EDGE_STYLES.concat(['width', 'height', 'position']);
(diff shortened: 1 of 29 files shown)


joneit:
@Dwaynekj @dcchuck @neilslinger Will be publishing these v2 fixes as 2.1.6 at EOD today if no questions or objections. Thanks to @nmichaud for consuming/testing! J.

src/dataModels/JSON.js
```diff
@@ -721,50 +725,73 @@ var JSON = DataModel.extend('dataModels.JSON', {
  * @this {dataModels.JSON}
  * @memberOf dataModels.JSON~
  */
-function getUnderlyingIndexesOfSelectedRows() {
-    var sourceIndexes = [],
```
Dwaynekj: Is there a deprecation warning for this?

joneit: "Inner" function (not exposed)

Dwaynekj: oh sorry, I should have expanded the view

Dwaynekj:
How do we reproduce
Fixed single-column-after-fixed-column(s) render issue

joneit:
@Dwaynekj Umm.... just what it says: When there is only one column after one or more fixed columns. So in the demo, we already have 1 fixed column, so just reduce the the total number of columns to 2 so that there is only one column after the fixed column, and you will see the renderer start to throw errors and hangs:
```js
grid.properties.columnIndexes = grid.properties.columnIndexes.slice(0, 2);
```
(You cannot use `.length = 2` here because the grid.properties.columnIndexes` array is generated by a getter. Interesting idea, though: We could override `length` on the return value to be another getter/setter.... hmmm!)

src/Hypergrid.js
```diff
@@ -889,6 +910,10 @@ var Hypergrid = Base.extend('Hypergrid', {
      * Called from renderer/index.js
      */
     deferredBehaviorChange: function() {
+        if (this.needsReindex) {
+            this.behavior.reindex();
```
Dwaynekj: Hey @joneit Can you explain what this feature is for?

joneit:
Hey @Dwaynekj `deferredBehaviorChange` makes `beahviorShapeChanged`, `behaviorStateChanged`, and `reindex` (synonym `applyAnalytics`) flag-bassed, to be executed once only at next render, rather than immediately. This means that these routines, like `repaint()`, can now be called repeatedly within a thread but will only ever be executed one time before the next render.

joneit:
I'm delaying **2.1.6** release for further testing. Should be done today though.....

joneit:
**2.1.6** looks good now.

Issue #708, opened 2018-04-09
Format cannot be overwritten in getCell


nclsclsn:
In renderer/index.js, format is set from grid properties and it cannot be dynamically overwritten via:

var cellRenderer = behavior.dataModel.getCell(config, config.renderer);

Is there a reason for it? If not, I believe it would make sense to make that property dynamic like any other.
Thanks!

joneit:
Hello Nicolas, Thanks for raising this issue...

Looks like a bug since (at least) **v1.3.0**… To clarify, by "grid properties" you actually meant cell || column || grid (and as of **v2.1.0**, row); and by "dynamic" you actually meant "as mutated by getCell"; and the errant line [from renderer/index.js] is actually:
```js
config.formatValue = grid.getFormatter(format);
```
which we shall change to:
```js
config.formatValue = grid.getFormatter(config.format);
```
I will propose this correction for a new patch **v2.1.7** to be released this week; and in **v3.0.0** (to be released in near future).

Jonathan

joneit:
I meant to add that, as a reminder, `getCell` is a programmatic remedy that is not as efficient as the "declarative" approach of setting column property (or cell property).

joneit:
Hello Nicolas @nclsclsn, Thanks for raising this issue...

Looks like a bug since (at least) v1.3.0… To clarify, by "grid properties" you actually meant cell || column || grid (and as of v2.1.0, row); and by "dynamic" you actually meant "as mutated by getCell"; and the errant line [from renderer/index.js] is actually:

config.formatValue = grid.getFormatter(format);
which we shall change to:

config.formatValue = grid.getFormatter(config.format);
I will propose this correction for a new patch v2.1.7 to be released this week; and in v3.0.0 (to be released in near future).

As a reminder, `getCell` hook is a programmatic approach that is not as clean/efficient as the "declarative" approach of setting cell or column `format` property.

Jonathan

joneit:
@nclsclsn PR #709 is up for review.

nclsclsn:
Thanks Jonathan for the quick feedback, clarification and fix.
I definitely favor the declarative approach when I can but formatting is data driven in my current use case i.e. some rows use formatting F1 when a field is set to X while others use formatting F2 when the same fied is different than X.
I believe I can't achieve such a conditional formatting with the declarative approach, can I?

PR #709, opened 2018-04-09
v2.1.7

joneit:
### Roadmap for this patch release (checked items completed):
* [x] `getCell`-mutated `config.format` now used (fixes bug reported in #708)
* [x] Added row properties optimization to `assignProps` to utilize it's grid props prototype
* [x] Optimized `assignProps` for these complex use cases:
   * column + cell
   * cell + column
   * cell + … + column
* [x] Expanded role of "fixed" vertical rule line. Previously, when there were no fixed columns, the vertical fixed rule line was not rendered at all. It now renders to right of tree column (when visible) else row number column (when visible).
* [x] Fixed rule lines now have a minimum thickness equal to that of the grid lines.
* [x] `boxSizing` (new grid property, analogous to CSS property [`box-sizing`](https://developer.mozilla.org/docs/Web/CSS/box-sizing)) has two possible settings:
   * `boxSizing === 'border-box'`.
      * Default setting in v2.1.7.
      * Cells are always rendered contiguously.
      * Grid rule lines are rendered inside cell at right and bottom edges.
      * Fixed are also rendered inside the cell, except for any portion of thickness exceeding grid line thickness, which is rendered to right of and below the cell, pushing cells to right rightward and cells below down.
   * `boxSizing === 'content-box'`:
      * Default setting in v3.0.0 (and `defaultRowHight` default will be one less to compensate).
      * Cells are spread apart horizontally by `gridLinesVWidth` pixels (rendered contiguously when `gridLinesVWidth === 0`).
      * Cells are spread apart vertically by `gridLinesHWidth` (rendered contiguously when `gridLinesHWidth === 0`).
   * `SimpleCell#paint` cell renderer now renders text 1 pixel higher up on odd cell heights.
   * `synchronizeScrollingBoundaries` method corrected for better position of last row/colun
      * Now respects cell gaps in calculating maximum scroll position so that the last row/column are positioned properly at bottom/right.
      * Now _always_ called from `grid.setData`
* [x] Fix column auto-sizing (#710)
* [x] Refine tree column auto-sizing:
   * Add `treeColumnAutosizing` property
   * Add `behavior.getTreeColumn` and `behavior.autoSizeTreeColumn` methods
* [x] Update package-lock.json (#711)

#### _Checked items committed. Unchecked items still in development._


src/Hypergrid.js
```diff
@@ -893,17 +901,27 @@ var Hypergrid = Base.extend('Hypergrid', {
      * @desc The dimensions of the grid data have changed. You've been notified.
      */
     behaviorShapeChanged: function() {
-        this.needsShapeChanged = true;
-        this.repaint();
+        if (!this.properties.repaintImmediately) {
+            this.needsShapeChanged = true;
+            this.canvas.repaint();
+        } if (this.divCanvas) {
```
nclsclsn: Could there be a else missing?

src/Hypergrid.js
```diff
@@ -893,17 +901,27 @@ var Hypergrid = Base.extend('Hypergrid', {
...
      * @memberOf Hypergrid#
      * @desc The dimensions of the grid data have changed. You've been notified.
      */
     behaviorStateChanged: function() {
-        this.needsStateChanged = true;
-        this.repaint();
+        if (!this.properties.repaintImmediately) {
+            this.needsStateChanged = true;
+            this.canvas.repaint();
+        } if (this.divCanvas) {
```
nclsclsn: Same here

joneit:
yikes! thx

PR #713, opened 2018-04-19
v2.1.8

joneit:
* fix deferred behavior change (good catch @nclsclsn!)
* upgraded to synonomous@2.1.2 so integer synonyms are not added to `columns` and `allColumns` arrays


joneit:
will push tonight or over weekend

PR #725, opened 2018-06-07
Disable alpha on canvas context.

mofojed:
Since the grid background is drawn anyway, the grid does not need to be transparent. This allows optimizations of the canvas for better performance: https://developer.mozilla.org/en-US/docs/Web/API/Canvas_API/Tutorial/Optimizing_canvas

In particular, it allows subpixel aliasing, which makes text look much crisper.


joneit:
@mofojed Thanks for the PR. My apologies for not responding earlier. I have to admit I was not aware of the `alpha` option. However, FYI, Hypergrid intentionally supports transparent/translucent grids, for example:
```js
grid.properties.backgroundColor = 'transparent'; grid.repaint();
grid.properties.backgroundColor = 'rgba(0,0,0,0)'; grid.repaint();
grid.properties.backgroundColor = 'rgba(200,0,0,.5)'; grid.repaint();
```

Also, I am confused by your statement that the text can be made crisper. Text is already clearly being anti-aliased. Setting alpha to true on graphics context creation made no visible difference to me.

Regarding canvas optimizations, I see no performance improvement whatsoever.

All that said, I admit the use case for a transparent or translucent grid is rare and so for that reason I am happy to add the feature as an option to v2.1.15 and v3.0.0 as a new grid property `canvasAlphaChannel: false`.

I was actually going to make the default `true` but testing (with Chrome v67) shows no performance improvement whatsoever. I shall proceed however, the assumption being that there are obscure use cases or other browsers for which it would make a difference.

Note that obviously this option needs to be set _prior to_ grid instantiation, so that means setting it globally:
```js
Hypergrid.defaults.canvasAlphaChannel = true
```
or in `state` option on instantiation:
```js
new Hypergrid({ state: { canvasAlphaChannel: true } })
```
Also, changing this value later will have no effect on the grid proper (since graphics context already created).

Please respond if you have any more information. In any case I am closing this PR but will attempt to cherry-pick your commit for inclusion in the repo history for this feature.

mofojed:
Thanks @joneit! `canvasAlphaChannel` looks good.

Yes, the text is already being anti-aliased, but it is not subpixel rendering anti-aliasing, which produces much clearer text. If you zoom in on a grid that is black text on a white background, the edges of the text have grays in them (normal anti-aliasing). Compare that to when you zoom in on text within a normal DOM element, and you'll see that the left edge is actually reddish, and the right side bluish. Wikipedia has a decent page on it: https://en.wikipedia.org/wiki/Subpixel_rendering

I'll try to set up a plunkr with the two grids side by side to demonstrate.

joneit:
@mofojed, I see it now. I've also updated the design: Rather than a _grid property,_ this will be an _instantiation option._ Please see PR #735 for complete description including a side-by-side comparison.

mofojed:
Awesome! The side-by-side comparison really makes it clear as well. It's a subtle change, but one you definitely notice when looking at grids of numbers for a while. Thanks!

PR #850, opened 2024-12-19
Vc 11716 reset snapshot after error cell being rendered

eclipse-rmo:
There is a lazy rendering logic that only renders a cell if the cell's state matches the previously rendered snapshot.

A small bug exists in this logic: when an error cell is rendered, the snapshot is not updated. Consequently, if the error cell returns to its previous value, the hypergrid assumes the cell has not changed and will not re-render it. This situation causes the error cell to stay.


eclipse-rmo:
mistake

